PR #10, opened 2014-11-07
Update index.html

sergejmueller:
Shortcut for _ssl on_
See http://nginx.org/en/docs/http/configuring_https_servers.html


gene1wood:
@sergejmueller Thanks for the PR. I implemented this slightly differently to accommodate older and newer nginx versions. You can see my tweak of your suggestion at a86cf4619d865c11beb1527c90abb22652dd71a6

PR #11, opened 2014-11-07
Added HSTS option to include sub-domains

neilstuartcraig:
Hi

I just added an option (via a checkbox) to include sub-domains in the HSTS HTTP header.
Adding in another few relatively minor updates too.

Cheers
Neil


gene1wood:
@neilstuartcraig Thanks for the PR. Couple questions.

Can you tell me about the HAProxy change of `Strict-Transport-Security:\` to `Strict-Transport-Security:\\\`. Is the first syntax not valid?

What's the rationale behind the removal of `SSLEngine on` from the apache config?

neilstuartcraig:
Hi Gene

Yes sure. The \ was;t coming out in the output for me, i think it was basically ending up escaping a space, hence the \\ which was the only way I could get a \ to appear in the output.

I didn’t mean to remove the SSLEngine On from the apache config, my apologies, i must have got a bit too keen on the ctrl+k.

Cheers
Neil

gene1wood:
I've incorporated most of what's in this PR but I'm going to hold off on adding the HSTS sub-domain checkbox so as to choose code simplicity of the feature. If others encounter a need for HSTS sub-domain, speak up here as I may have an inaccurate sense of the number of people who would need this functionality.

lgarron:
Nearly all new submissions to the HSTS preload list have `include_subdomains: true`.

https://code.google.com/p/chromium/codesearch#chromium/src/net/http/transport_security_state_static.json

Assuming that your site can handle it (which unfortunately must be evaluated on a case-by-case basis for legacy sites), it's a really good idea to set it.

lgarron:
Actually, it's literally _all_ domains on the preload list (except Yahoo, which has a bunch of entries for explicit subdomains).

jvehent:
`include_subdomains` is as useful as it is dangerous, especially for Mozilla where we have hundreds of subdomains hosted in various places, not always under standard configurations.
I think it's good to have it available as an option, but I would argue that this option must be unset by default.

PR #24, opened 2015-02-16
Update session timeout on nginx

lcamacho:
Per recommendation of Ilya Grigorik, session timeout should be 1 day https://docs.google.com/presentation/d/1BH9DI1XlmukCzU2i8OvxLIfgQf_aGlZgZyvWDSyYyzs/present?slide=id.g41fa27ab0_4_18


gene1wood:
@lcamacho thanks for the suggestion, good idea. I've added your change.

PR #46, opened 2015-04-13
Adding new semver test for OpenSSL grammar and code to disable `modern` profile for old versions of OpenSSL

### LINKED ISSUE #26: No modern configuration available with openSSL<1.0.1
Hi,
openSSL 0.9.8 and 1.0.0 don't support TLSv1.1 or TLSv1.2 (only appear with 1.0.1 cf http://en.wikipedia.org/wiki/OpenSSL#Major_version_releases), so the "Modern" configuration gave by the generator can't be used.
it may be a boog idea to replace the configuration by a warning/error message like "openSSL version does not support the requirement for a modern configuration"

gene1wood:
Setting more common default values for Apache and OpenSSL
Resolves #26

### DIFF
--- a/ssl-config-generator/index.html
+++ b/ssl-config-generator/index.html
@@ -31,6 +31,7 @@
     <script>
         // https://gist.github.com/cowboy/566233
         var isSemVer=(function(){var a=/^(<|>|[=!<>]=)?\s*(\d+(?:\.\d+){0,2})([a-z][a-z0-9\-]*)?$/i;function b(e,c){var d=(e+"").match(a);return d?(c?(d[1]||"=="):"")+'"'+(d[2]+".0.0").match(/\d+(?:\.\d+){0,2}/)[0].replace(/(?:^|\.)(\d+)/g,function(g,f){return Array(9-f.length).join(0)+f;})+(d[3]||"~")+'"':(c?"==0":1);}return function(e){e=b(e);for(var c,d=1;c=arguments[d++];){if(!(new Function("return "+e+b(c,1)))()){return false;}}return true;};})();
+        var isOpenSSLSemVer=(function(){var a=/^(<|>|[=!<>]=)?\s*(\d+(?:\.\d+){0,2})([a-z][a-z0-9\-]*)?$/i;function b(e,c){var d=(e+"").match(a);return d?(c?(d[1]||"=="):"")+'"'+(d[2]+".0.0").match(/\d+(?:\.\d+){0,2}/)[0].replace(/(?:^|\.)(\d+)/g,function(g,f){return Array(9-f.length).join(0)+f;})+(d[3]||"")+'"':(c?"==0":1);}return function(e){e=b(e);for(var c,d=1;c=arguments[d++];){if(!(new Function("return "+e+b(c,1)))()){return false;}}return true;};})();
     </script>
     <script>
         if (window.location.protocol != "https:")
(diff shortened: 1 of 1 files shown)


ssl-config-generator/index.html
```diff
@@ -31,6 +31,7 @@
     <script>
         // https://gist.github.com/cowboy/566233
         var isSemVer=(function(){var a=/^(<|>|[=!<>]=)?\s*(\d+(?:\.\d+){0,2})([a-z][a-z0-9\-]*)?$/i;function b(e,c){var d=(e+"").match(a);return d?(c?(d[1]||"=="):"")+'"'+(d[2]+".0.0").match(/\d+(?:\.\d+){0,2}/)[0].replace(/(?:^|\.)(\d+)/g,function(g,f){return Array(9-f.length).join(0)+f;})+(d[3]||"~")+'"':(c?"==0":1);}return function(e){e=b(e);for(var c,d=1;c=arguments[d++];){if(!(new Function("return "+e+b(c,1)))()){return false;}}return true;};})();
+        var isOpenSSLSemVer=(function(){var a=/^(<|>|[=!<>]=)?\s*(\d+(?:\.\d+){0,2})([a-z][a-z0-9\-]*)?$/i;function b(e,c){var d=(e+"").match(a);return d?(c?(d[1]||"=="):"")+'"'+(d[2]+".0.0").match(/\d+(?:\.\d+){0,2}/)[0].replace(/(?:^|\.)(\d+)/g,function(g,f){return Array(9-f.length).join(0)+f;})+(d[3]||"")+'"':(c?"==0":1);}return function(e){e=b(e);for(var c,d=1;c=arguments[d++];){if(!(new Function("return "+e+b(c,1)))()){return false;}}return true;};})();
```
jvehent, line 35: I must be missing something... why do you need both `isSemVer` and `isOpenSSLSemVer` ? Can't you just use `isSemVer` all the time?

Also, slightly unrelated, but it'd be nice to use the non-minified version instead. It cost a tiny bit more in transfer but way easier to read.

ssl-config-generator/index.html
```diff
@@ -214,6 +215,17 @@
                     cipherSuites: cipherSuites[data.securityProfile],
                     clientList: clientList[data.securityProfile]
                 });
+
+                if (isOpenSSLSemVer(data.opensslVersion, "<1.0.1") && data.securityProfile == "modern") {
+                	$("#security-profile-list input#intermediate").prop( "checked", true );
+                	data.securityProfile = "intermediate";
+                	$("#security-profile-list input#modern").prop("disabled", true);
+                	$("#security-profile-list label[for=modern]").attr("title", "Modern profile is not possible with OpenSSL <1.0.1");
```
jvehent: Could the message indicate `why` modern is not possible? Like: is tlsv1.{1,2} support missing? no cipher overlap? etc...

ssl-config-generator/index.html
```diff
@@ -262,10 +274,10 @@
     </div>
     <div style="width 100px; float:left; padding:1em;">
         <label for="server-version">Server Version</label>
-        <input id="server-version" type="text" maxlength="15" value="1.0" />
+        <input id="server-version" type="text" maxlength="15" value="2.2.15" />
         <br />
         <label for="openssl-version">OpenSSL Version</label>
-        <input id="openssl-version" type="text" maxlength="15" value="0.9.8h" />
+        <input id="openssl-version" type="text" maxlength="15" value="1.0.1e" />
```
jvehent, line 281: r+. rhel6 ship with 1.0.1e so that seems appropriate as a default

ssl-config-generator/index.html
gene1wood, line 35: The reason for both functions is that when comparing everything other than OpenSSL, (for example Apache or Nginx versions), we should use the traditional semver rules (1.0.1 > 1.0.1pre3). In the case of OpenSSL that don't follow semver we need the custom function (1.0.1 < 1.0.1e ).

My thought with the minified / non-minified code was that by linking to the non-minified code in the comment above it, people could look at it if they needed. Ideally the semver logic should not be in much flux and shouldn't require examination. What do you think?

jvehent, line 35: That makes sense, I'd suggest posting your justification for 2 functions in the code, I didn't notice the diff when looking at them.

since 2 functions are required, it would look weird to store the expanded versions in the code. Let's hope that gist is never deleted :)

r+

gene1wood:
@jvehent Good ideas on a comment for the rationale for two semver versions as well as more detail in the message about why modern isn't available.

PR #47, opened 2015-04-13
Max DH Key size is constrained by Java support

### LINKED ISSUE #31: Invalid DH param for Java 6
We've been testing the generator and found that the `tune.ssl.default-dh-param` setting recommendation is incorrect when trying to support Java 6 (the old config). Java 6 has a maximum acceptable size of 1024.

gene1wood:
Looks like Java 7 (intermediate) and Java 6 (old) both only support `1024` bit keys

http://cbonte.github.io/haproxy-dconv/configuration-1.5.html#tune.ssl.default-dh-param

Fixes #31

### DIFF
--- a/ssl-config-generator/index.html
+++ b/ssl-config-generator/index.html
@@ -96,7 +96,7 @@ server {
 <pre style="visibility: {{visibility}};">
 global
     # set default parameters to the {{securityProfile}} configuration
-    tune.ssl.default-dh-param 2048
+    tune.ssl.default-dh-param {{maxDHKeySize}}
     ssl-default-bind-ciphers {{cipherSuites}}
 
 frontend ft_test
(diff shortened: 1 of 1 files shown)


gene1wood:
@jvehent is looking into confirming that Java 7 does not support 2048 bit

jvehent:
Confirmed, Java 7 fails to perform a handshake when the DH parameter is set to 2048.

```
[javax.net.ssl.SSLException]: java.lang.RuntimeException: Could not generate DH keypair
```

I have changed the `intermediate` configuration to require 1024 bits DHE parameters, since this is what we use everywhere anyway: https://wiki.mozilla.org/Security/Server_Side_TLS#Intermediate_compatibility_.28default.29

Issue #48, opened 2015-04-13
Explain how to generate dhparam


thefinn93:
Both the nginx and HAProxy configs include a path to a `dhparam.pem`, but there is no explanation for how to generate it. Googling indicates it's pretty simple (`openssl dhparam -out /path/to/dhparam.pem 2048`) but it requires an extra search that seems like it shouldn't be needed. I suggest adding a comment above the line that mentions it indicating how to generate it.

jvehent:
A number of servers provide default parameters out of the box, but we could link to the wiki page https://wiki.mozilla.org/Security/Server_Side_TLS#DHE_handshake_and_dhparam

gene1wood:
@jvehent What do you recommend? Should we add a comment to the nginx and haproxy configs pointing to that wiki link? Should we add instructions on how to generate a dhparam file to the wiki and add a comment to the config linking to it? Something else?

jvehent:
I'd recommend linking to the wiki for knowledge dissemination. The command is simple, but understanding how the DHE parameter is used warrants some explanations that the wiki already has.

ghost:
@jvehent @gene1wood Could we leave this as a Good First Bug and have @thefinn93 submit a pull request should he be interested?

gene1wood:
@bkerensa Sure.
@thefinn93 comment here if that sounds good. If not or if I don't hear from you I'll take care of it.

thefinn93:
I'm not entirely sure how to add/test that, I guess just edit [this line](https://github.com/mozilla/server-side-tls/blob/gh-pages/ssl-config-generator/index.html#L273)?

gene1wood:
@thefinn93 Looks like what @jvehent suggested was to add a comment in the produced config file that is a link to the wiki page describing more information about dhparam. Do probably you'll want to edit the config payload for haproxy

thefinn93:
Alright, I assume the nginx payload too? That line I linked seems to be more nginx specific, didn't realize it was in the HA Proxy payload

jvehent:
I'm not sure about nginx, but haproxy will generate a diffie hellman parameter on its own when `tune.ssl.default-dh-param` is set. So, is it really necessary to point people to manual instructions when the tool takes care of that for you?

thefinn93:
nginx will not do that, hence this thread. I have no idea about haproxy. Should I remove the comment from the HA Proxy config?

nbibler:
[...]

PR #57, opened 2015-04-16
Link to the dhparam generation info on the wiki

thefinn93:
Follow up to https://github.com/mozilla/server-side-tls/issues/48

### DIFF
--- a/ssl-config-generator/index.html
+++ b/ssl-config-generator/index.html
@@ -106,6 +106,8 @@ global
 
 frontend ft_test
     mode    http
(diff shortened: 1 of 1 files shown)


gene1wood:
@jvehent @thefinn93 asks "Should I remove the comment from the HA Proxy config?"

jvehent:
@thefinn93 : Yes, it would be preferable to remove the comment from the haproxy config and let it take care of the generation. The comment can stay for nginx.

jvehent:
I'm rewriting parts of the generator and will include theses changes differently.

PR #58, opened 2015-04-17
Change HSTS header setting in Apache

ghost:
Is it better to use "Header always set Strict-Transport-Security" than "Header always add Strict-Transport-Security", in order to prevent adding duplicate HSTS headers?

The doc explains the "add" flag as follows, "the response header is added to the existing set of headers, even if this header already exists. This can result in two (or more) headers having the same name. This can lead to unforeseen consequences, and in general set, append or merge should be used instead." [1]
[1] https://httpd.apache.org/docs/2.2/mod/mod_headers.html#header

### DIFF
--- a/ssl-config-generator/index.html
+++ b/ssl-config-generator/index.html
@@ -316,7 +316,7 @@ frontend ft_test
                     }
                     if (data.hstsEnabled == "true") {
                         data.hsts = '\n    # HSTS (mod_headers is required) (15768000 seconds = 6 months)' + '\n' +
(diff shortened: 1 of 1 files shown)


jvehent:
Yes, `set` is better than `add`. Thanks for the patch.

ghost:
@jvehent I don't have a Mozilla Wiki account, so could you also update the Wiki page https://wiki.mozilla.org/Security/Server_Side_TLS#Apache?

change
`Header add Strict-Transport-Security "max-age=15768000"`
to
`Header always set Strict-Transport-Security "max-age=15768000"`

Thanks!

PR #70, opened 2015-06-07
add notes about pre-defined DH groups, DHE support in clients

rgacogne:


jvehent:
Looking good! Thanks for taking my comments, and thanks again for writing this! :)

PR #73, opened 2015-06-10
Change version history to proper table

april:
Oops, this pull request should have the full history.  :)


jvehent:
I like it! It's super clean. Merging now, and with thanks :)

PR #76, opened 2015-06-22
Typos (links)

drwetter:
https://www.mediawiki.org/wiki/Help:Links#External_links


jvehent:
Thanks for the corrections :)

PR #79, opened 2015-07-09
Bump maxDHKeySize to 2048 for intermediate profile

meineerde:
In version 3.6 of the Server Side TLS document, the DH size in the intermediate was bumped to 2048 bits. With ECDHE available in the cipher list, even Java 7 should be able to connect. See https://wiki.mozilla.org/Security/Server_Side_TLS#DHE_and_Java for more information.


jvehent:
Thanks for the patch! I changed it on the wiki and forgot to update the conf generator :)

PR #80, opened 2015-07-09
Disable TLS tickets by default

meineerde:
By default, all servers enable TLS tickets. However, the keys are only renewed on server restart, leasing to the unfortunate situation that the secret key for the tickets doesn't change for a long time which effectively destroys the perfect-forward-secrecy guarantee. While all servers allow to specify a file which contains the key on disk, it is generally recommended to not use this as this allows the key to leak under effectively the same situations the private key could leak, which again defeats the purpose of PFS.

The use of server-stored sessions (identified by session ids) is not affected by this and is always safe.

Because of this, The Document™ specifies at https://wiki.mozilla.org/Security/Server_Side_TLS#TLS_tickets_.28RFC_5077.29 that support for TLS tickets should be disabled if possible. This patch adds the relevant config options.

Note that Apache only supports this setting since 2.2.30 when used with OpenSSL >= 0.9.8f - http://httpd.apache.org/docs/2.2/mod/mod_ssl.html#sslsessiontickets

Some versions of HAProxy (i.e >= 1.6-dev2) allow to set the key via the stats socket. This however requires custom scripts and good maintenance on the side of the operator. A safe default in any case is to disable session tickets.

### DIFF
--- a/ssl-config-generator/index.html
+++ b/ssl-config-generator/index.html
@@ -54,6 +54,7 @@ server {
     ssl_certificate_key /path/to/private_key;
     ssl_session_timeout 1d;
     ssl_session_cache shared:SSL:50m;
+    ssl_session_tickets off;
 {{dhparam}}
 
     # {{securityProfile}} configuration. tweak to your needs.
@@ -85,6 +86,7 @@ server {
     SSLProtocol             {{sslProtocols}}
     SSLCipherSuite          {{cipherSuites}}
     SSLHonorCipherOrder     on
+{{sslSessionTickets}}
 {{compression}}
 {{ocspStapling}}
 {{hsts}}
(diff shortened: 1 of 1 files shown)


jvehent:
What happens is the option is present in apache < 2.2.30? If it breaks apache, then it should be put into a condition to the version requested in the generator.

meineerde:
I updated the pull request with a conditional for Apache. Unfortunately, I'm not fully sure how to run the generator locally so I can't test it. But as it works very similar to other places in the file, I;m rather confident that this should do it :)

jvehent:
Looks good to me. Sorry for the long review, and thanks for the patch.

jrchamp:
Shouldn't the section about enabling TLS tickets in F5 be removed or tell you instead how to disable it?

meineerde:
@jrchamp I don't know how exactly F5 implements this feature on their loadbalancers. The point of disabling it by default is that most products don't have a secure default configuration for TLS tickets and securing it requires some scripting and generally requires some understanding of TLS and physical security.

If F5 implements this securely, i.e. if they regularly generate new ticket keys and ensure that these keys are never stored on persistent memory, it is safe to enable.

jrchamp:
@meineerde Thank you for the extremely fast response! I'll follow up with them.

PR #84, opened 2015-08-16
Remove a dead in-page link for Nginx

malcolmr:
The additional information for Nginx was removed in
https://wiki.mozilla.org/index.php?title=Security/Server_Side_TLS&diff=990137&oldid=983316.


jvehent:
Thanks for the patch!

PR #85, opened 2015-08-20
Add command to generate dhparam.pem

wolever:
Adds the OpenSSL command to generate the dhparam.pem file


jvehent:
See #57, I'm rewriting some section of the generator, and will add generic instructions for all DH handling.

PR #89, opened 2015-08-28
Add ECDHE-3DES ciphers to intermediate level

jvehent:


april:
Yep, looks right to me.  Adds these two upside-down-world ciphers only:

```
0xC0,0x12 - ECDHE-RSA-DES-CBC3-SHA  SSLv3 Kx=ECDH     Au=RSA  Enc=3DES(168) Mac=SHA1
0xC0,0x08 - ECDHE-ECDSA-DES-CBC3-SHA SSLv3 Kx=ECDH     Au=ECDSA Enc=3DES(168) Mac=SHA1
```

PR #94, opened 2015-09-23
15724800 -> 15768000, to bring in line with generator

april:
The difference between 182 and 182.5 days.


jvehent:
those damn 43200 seconds!

r+ :+1:

PR #95, opened 2015-09-24
Lighttpd support + tweaks

synapt:
Added in some lighttpd config support to the ssl config generator, plus a few other tweaks on the page which were admittedly just some nagging nitpicks to me, lol.

Also updated application versions for the auto complete.


ssl-config-generator/index.html
```diff
@@ -200,7 +236,8 @@
                 sslProtocols: {
                     apache: 'all -SSLv2 -SSLv3',
                     nginx: 'TLSv1 TLSv1.1 TLSv1.2',
-                    haproxy: 'ssl no-sslv3'
+                    haproxy: 'ssl no-sslv3',
+                    lighttpd: '    ssl.use-sslv2 = "disable"\n    ssl.use-sslv3 = "enable"'
```
jvehent: `sslv3` should be disabled at intermediate level

ssl-config-generator/index.html
```diff
@@ -124,6 +124,41 @@
...
+    ssl.engine   = "enable"
+    ssl.disable-client-renegotiation = "enable"
+
+    # pemfile is cert+privkey, ca-file is the intermediate chain in one file
+    ssl.pemfile               = "/path/to/signed_cert_plus_private_key.pem"
+    ssl.ca-file               = "/path/to/intermediate_certificate.pem"
+    # for DH/DHE ciphers, dhparam should be >= 2048-bit
+    ssl.dh-file               = "/path/to/dhparam.pem" 
+    # ECDH/ECDHE ciphers curve strength (see `openssl ecparam -list_curves`)
+    ssl.ec-curve              = "secp384r1"
```
jvehent, line 142: Can this take a list? if not, `p256` is more common

ssl-config-generator/index.html
```diff
@@ -124,6 +124,41 @@
...
+    # Compression is by default off at compile-time, but use if needed
+    # ssl.use-compression     = "disable"
```
jvehent, line 144: if it's disabled at compile time, will enabling it here do anything?

jvehent:
Looks interesting! Couple comments in there, but it should be fairly easy to merge.

ssl-config-generator/index.html
synapt, line 142: Just a single value, I admittedly thought 384r1 was the more common one.

ssl-config-generator/index.html
synapt, line 144: Should just throw an error into the log file along the lines of ssl.use-compression being either unavailable or an unknown directive if someone uncomments that w/ compression disabled, I was just mainly considering possibly goofed up packaged installs of lighty that may have it enabled.

ssl-config-generator/index.html
synapt: Ah, missed that one, sorry :(

ssl-config-generator/index.html
jvehent, line 142: After further discussions, it appears that p384 is the way to go, so you should leave that in place.

synapt, line 142: Ah alright, will do :P

jvehent:
r+! Thanks for the patch!

PR #98, opened 2015-11-03
Updated checking version of Apache and hide unsupported directive

gaeulbyul:
In some version of Apache doesn't support SSLSessionTickets directive. So I updated checking version and hiding unsupported directive code.

In addition, also updated SSLCompression that require OpenSSL >= 0.9.8, and availability of Apache >= 2.2.24


jvehent:
r+. Thanks for the patch!

PR #100, opened 2015-11-03
Attempt to fix SSLSessionTickets bug

### LINKED ISSUE #99: SSLSessionTickets not supported in Apache < 2.4.11
The generator outputs the `SSLSessionTickets` option even for Apache versions that don't support it.

Apache manual:
http://httpd.apache.org/docs/2.4/mod/mod_ssl.html#sslsessiontickets

Example generating bad config:
https://mozilla.github.io/server-side-tls/ssl-config-generator/?server=apache-2.4.7&openssl=1.0.1f

steffann:
Untested, but should resolve #99


jvehent:
I merged #98 which fixes the same issue, so I can't merge this patch, but thanks for submitting it anyway!

PR #102, opened 2015-11-05
Move Apache server config outside VirtualHost

jrchamp:
Rather than set the SSL default configuration inside of each virtual host, set it at the server level. Only virtual host specific customizations/overrides should be inside of the VirtualHost.

Thoughts?

### DIFF
--- a/ssl-config-generator/index.html
+++ b/ssl-config-generator/index.html
@@ -82,16 +82,17 @@ server {
     SSLCertificateKeyFile   /path/to/private/key
     SSLCACertificateFile    /path/to/all_ca_certs
 
-    # {{securityProfile}} configuration, tweak to your needs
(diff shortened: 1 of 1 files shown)


jvehent:
The code snippet this generator provides is very vhost specific. I think if it were to provide global parameters, it should split the rendering into two snippets: one for global that will likely go into httpd.conf, and one for vhost that will likely go into a vhost conf file.

jvehent:
I just realized we already have global parameters showing out of the vhost snippet, so maybe this is fine as it is :)
Could you rebase your branch?

jrchamp:
Rebased. :)

jvehent:
Let's try it out and see if it works for everyone :)
Thanks for the patch!

PR #103, opened 2015-11-18
nginx: ssl_session_tickets appeared first in 1.5.9

vthriller:
See http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_session_tickets

Also, replicate OpenSSL version constrain from 'case "apache"'.


jvehent:
r+. Thanks for the patch!

PR #105, opened 2015-11-21
Trigger re-render on checkbox toggle

dittos:
Toggling HSTS checkbox didn't trigger configuration to be re-rendered.


jvehent:
Good catch! Thanks for the patch.

PR #108, opened 2016-01-12
Add nginx HTTPS redirect with HSTS, HTTP/2, and IPv6

corburn:
Hello, would the following changes be appropriate to include with the nginx generator?
- Always listen on both IPv4 and IPv6
- Enable HTTP/2 with nginx >=1.9.5
- Include an HTTP to HTTPS redirect when using HSTS

```
server {
    listen 80 default_server;
    listen [::]:80 default_server;

    # HSTS (ngx_http_headers_module is required) (15768000 seconds = 6 months)
    add_header Strict-Transport-Security max-age=15768000;

    # Redirect all HTTP requests to HTTPS with a 301 Moved Permanently response.
    return 301 https://$host$request_uri;
}

server {
    listen 443 ssl http2;
    listen [::]:443 ssl http2;
    [...]
}
```

On another note, with the public beta of LetsEncrypt.org, I'm curious if it would be appropriate to include an example of using the using the webroot certificate renewal:

```
server {
    listen 80 default_server;
    listen [::]:80 default_server;

    # HSTS (ngx_http_headers_module is required) (15768000 seconds = 6 months)
    add_header Strict-Transport-Security max-age=15768000;

    # LetsEncrypt.org TLS certificate renewal
    location ~ .well-known/acme-challenge/ {
        root /srv/letsencrypt;
    }

    # Redirect all HTTP requests to HTTPS with a 301 Moved Permanently response.
    location / {
        return 301 https://$host$request_uri;
    }
}
```

Finally, while I was reading how to configure nginx I didn't see too many examples that included headers such as in the [OWASP list of useful HTTP headers](https://www.owasp.org/index.php/List_of_useful_HTTP_headers). Should they be included in the generator or linked at the bottom of the page?

### DIFF
--- a/ssl-config-generator/index.html
+++ b/ssl-config-generator/index.html
@@ -46,6 +46,16 @@
 <h2>{{server}} {{serverVersion}} | {{securityProfile}} profile | OpenSSL {{opensslVersion}} | <a href="?{{queryString}}">link</a></h2>
 <p>Oldest compatible clients : {{clientList}}</p>
 <pre>
+{{#if hstsEnabled}}
+server {
+    listen 80 default_server;
+    listen [::]:80 default_server;
+    
+    # Redirect all HTTP requests to HTTPS with a 301 Moved Permanently response.
+    return 301 https://$host$request_uri;
+}
+
+{{/if}}
 server {
 {{listen}}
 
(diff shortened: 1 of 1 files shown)


tomato42:
+1 to IPv6 and HTTP2
-1 to HSTS by default and redirect

the problem with HSTS and other headers recommended by OWASP is that they can interact (badly) with some things that the applications are already doing and expecting to work. Enabling them should be a concious decision by the administrator and not a side-effect of applying a TLS policy.

corburn:
The changes in the pull request do not include HSTS and the redirect by default. The redirect is only included if the HSTS Enabled checkbox is checked
- [x] HSTS Enabled

The HSTS header in the redirect is based on one of the requirements for a website to be included in the browser [HSTS preload list](https://hstspreload.appspot.com/)

> If you are serving an additional redirect from your HTTPS site, that redirect must still have the HSTS header (not the page it redirects to).

tomato42:
aah, sorry, didn't notice `#if hstsEnabled`, ok then

lgarron:
Also note that you should not send the HSTS header over plain HTTP.

Browsers will ignore it, and it just makes your site look like it doesn't know what it's doing (and placing it in a tool like this would spread misconceptions).

corburn:
Thank you for the feedback tomato42 and Igarron. The HSTS header has been removed from the HTTP redirect. I did not see where in the spec it says the HSTS header should not be sent over HTTP, but checking with chrome://net-internals/#hsts showed it was only set when the header was sent over HTTPS.

lgarron:
> I did not see where in the spec it says the HSTS header should not be sent over HTTP,

It's a bit hard to find, but it's there:

> An HSTS Host MUST NOT include the STS header field in HTTP responses conveyed over non-secure transport. (https://tools.ietf.org/html/rfc6797#section-7.2)

mozillazg:
+1 for HTTP 2

jvehent:
This looks good, I'm merging it. Any chance you'd have something similar for Apache?
And thanks for the patch :+1:

corburn:
Thanks @jvehent. I look forward to someone who is more familiar with Apache submitting the next patch.

PR #126, opened 2016-03-02
Better defaults for HAProxy

yajo:
Redirects with 301 HTTP to HTTPS and adds cipher and options by default to all SSL binds and servers.


jvehent:
That looks good, thanks for the patch. Do you know if those parameters are available on all version of haproxy >= 1.5 ?

yajo:
Yes.

Actually with this settings you get the HSTS header only in the HTTPS site, since redirection is performed before header injection. But it's still better than it was.

lgarron:
> you get the HSTS header only in the HTTPS site

Reminder: that is the correct behaviour.

https://tools.ietf.org/html/rfc6797#section-7.2

yajo:
Oh nice

PR #133, opened 2016-03-19
Comment SSLCACertificateFile

### LINKED ISSUE #132: Use of `SSLCACertificateFile` not very clear
Hi guys,

If you search on Google for `SSLCACertificateFile`, you'll get many, _many_ results where `SSLCACertificateFile` is abused for sending a intermediate certificate.

If we look at the [docs](https://httpd.apache.org/docs/2.4/mod/mod_ssl.html#sslcacertificatefile) however, it is _clearly_ stated `SSLCACertificateFile` is used for **client authentication**.

Furthermore, if someone would interpret this as it should contain _all_ CA certificates (for example, `/etc/ssl/certs/`) and one would leave `SSLCADNRequestFile` or `SSLCADNRequestPath` empty ánd would set up client authentication somewhere, the server will send _all_ CA names to the client.

That can't be proper behaviour.

My suggestion is to add a comment above `SSLCACertificateFile` mentioning the fact it should only be used for _client authentication purposes_.

osirisinferi:
Fixes #132 

Note: this directive (add CA certificates for client authentication use) is only used in the Apache configuration, never in the other server software. Perhaps decide to delete it altogether?

### DIFF
--- a/ssl-config-generator/index.html
+++ b/ssl-config-generator/index.html
@@ -90,7 +90,9 @@ server {
     SSLEngine on
 {{certFile}}
     SSLCertificateKeyFile   /path/to/private/key
-    SSLCACertificateFile    /path/to/all_ca_certs
+
+    # Uncomment the following directive when using client certificate authentication
+    #SSLCACertificateFile    /path/to/ca_certs_for_client_authentication
 
 {{hsts}}
     ...


jvehent:
Looks good. Thanks!

PR #81, opened 2015-07-30
Apache 2.2.23 and up also support TLS 1.1 and 1.2

Gillingham:
Closes #62

PR #137, opened 2016-04-05
Update oldApache TLS support message

jrchamp:
Apache 2.2.23 and newer support TLS 1.1 and 1.2; update text to match commits from #81


jvehent:
Looks good. Thanks!

PR #141, opened 2016-04-21
use preferred haproxy header manipulation method

rremer:
Love this tool, but a minor nitpick on haproxy syntax used for setting the hsts max age. From the haproxy documentation:
"""
 Using "rspadd"/"rspdel"/"rsprep" to manipulate request headers is discouraged
  in newer versions (>= 1.5). But if you need to use regular expression to
  delete headers, you can still use "rspdel". Also please use
  "http-response deny" instead of "rspdeny".
"""

This actually doesn't mention a subtle but important difference in the syntax though: with 'rspadd' you will not get this setting if the client already set an hsts header. With 'http-response set-header' the server actually asserts the header is what the config says, even if the client were to say otherwise on every request.


jvehent:
Looks good. Sorry for the long delay!

Issue #144, opened 2016-06-08
Enable the new GitHub pages "Enforce HTTPS" option


edmorley:
See:
https://help.github.com/articles/securing-your-github-pages-site-with-https/

Thanks :-)

jvehent:
It's done automatically by github.

edmorley:
I'm not sure I follow?

This is a new feature that wasn't supported by GitHub until today. Existing github pages need the option enabling (only newly created gh-pages sites will have it enabled by default), hence filing this issue :-)

gene1wood:
@edmorley I've confirmed that the setting is currently checked (I didn't check it).

jvehent:
I checked it. But even unchecked, I get redirected to HTTPS. Not sure why, since there is no HSTS, but that's what the server does.

edmorley:
> I checked it. But even unchecked, I get redirected to HTTPS. Not sure why, since there is no HSTS, but that's what the server does.

That's the client-side redirect, which can now be removed (see #14).

edmorley:
I've opened #145 to remove the client-side redirect.

jvehent:
Awesome! Thanks!

PR #145, opened 2016-06-08
Remove client-side redirect now GitHub pages support enforcing HTTPS

edmorley:
GitHub pages now support enforcing HTTPS (which has been enabled in #144), so the client-side redirect can now be removed. See:
https://help.github.com/articles/securing-your-github-pages-site-with-https/


jvehent:
:+1:

Issue #146, opened 2016-06-20
no nginx-1.10 support


HalisCz:
There is no support for current stable nginx version (1.10 according to [nginx PPA archive](https://launchpad.net/~nginx/+archive/ubuntu/stable/+packages))

nbibler:
"No support," I'm guessing means it doesn't auto-fill with the current version of nginx in the text field?

I believe the configuration that's currently generated for nginx correctly works with and configures nginx 1.10. I haven't seen any issues with it in production.

Are you seeing differently?

ghost:
Now we have nginx 1.11.3 - http://nginx.org/en/CHANGES :-)

HalisCz:
> I believe the configuration that's currently generated for nginx correctly works with and configures nginx 1.10. I haven't seen any issues with it in production.
> 
> Are you seeing differently?

To be honest, I didn't try.

april:
Yeah, I think this is just a comment thing; there haven't been any changes in the nginx conf in some time that would affect the functioning of the actual configuration.

april:
The new version of the generator doesn't keep up to date with software versions, due to this issue:
https://mozilla.github.io/ssl-config-generator/

PR #151, opened 2016-07-28
Add nginx versions 1.9.6 to 1.10.1

nbibler:
This updates the list of stable releases of nginx and addresses (I think) #146.


jvehent:
Simple enough! Thanks for the patch.

Issue #82, opened 2015-07-31
Add support for Tomcat 7 + 8


fusrodave:
Tomcat is not a currently listed server, so there is no profile for a HTTPS connector. Would be great to have one added so Coyote can be used to serve HTTPS in a modern, secure manner.

jvehent:
I know for certain we never terminate HTTPS on anything java at Mozilla, so I'm afraid I don't have code to contribute. Do you have anything?

gene1wood:
Ya, @cuzzasoft a PR is welcome for Tomcat

jauderho:
This is somewhat difficult given that there are different Java versions that can plug into different versions of Tomcat. Further complicating the issue is minor revisions of Tomcat have steadily added additional features such as XFO etc.

How would that be handled?

I do have some notes on possible combinations that I'm happy to add but it's only for Tomcat 7 + Java 7 at this time.

stevespringett:
I'm also interested in this subject, but I don't think there will be all that many combinations if we only concentrate on Java versions that are not end of life (currently only Java 1.8), and the latest Tomcat release + point release (currently only Tomcat 8.0 and 8.5).

stevespringett:
The following config unveils the green lock in Chrome. This config is Tomcat-specific and works with Tomcat 8 on Java 8. This config does not specify precisely what ciphers to use.

tomcat/conf/server.xml

``` xml
<Connector port="8080" protocol="HTTP/1.1"
    connectionTimeout="20000"
    redirectPort="443" />

<Connector port="8443" protocol="org.apache.coyote.http11.Http11NioProtocol"
    maxThreads="150" SSLEnabled="true" scheme="https" secure="true"
    clientAuth="false" sslProtocol="TLS" sslEnabledProtocols="TLSv1.2" 
    keystoreFile="/path/to/keystore.jks" keyAlias="server" keystorePass="jkspassword"/>
```

tomcat/conf/web.xml

``` xml
<filter>
    <filter-name>httpHeaderSecurity</filter-name>
    <filter-class>org.apache.catalina.filters.HttpHeaderSecurityFilter</filter-class>
    <init-param>
        <param-name>hstsEnabled</param-name>
        <param-value>true</param-value>
    </init-param>
    <init-param>
        <param-name>hstsMaxAgeSeconds</param-name>
        <param-value>31536000</param-value>
    </init-param>
    <init-param>
        <param-name>hstsIncludeSubDomains</param-name>
[...]

PR #153, opened 2016-08-03
Adding Tomcat and JSSE ciphers

stevespringett:
Implementing Tomcat support using BIO/NIO support and JSSE. This implementation does not require OpenSSL via native (APR). It relies solely on the Java runtime itself.
#82


gene1wood:
Steve, thanks so much for this PR, it's great. I extended your PR in a new PR #161 which uses your code but generates the iana cipher list instead of embedding it statically. I think that'll be more maintainable over time. I'll get it reviewed and merged shortly. Thank you again for 02fd894

PR #157, opened 2016-08-24
Chrome 53 is removing DH

baloo:
If DH is present in the cipher list and any previous cipher in the list
doesn't match with the chrome internal cipher list, the connection will
result in a ERR_SSL_OBSOLETE_CIPHER connection error. The only fix we
found has been to totally remove DHE from the cipherlist.

Reference:
  https://codereview.chromium.org/2056343006
  https://www.chromestatus.com/features/5128908798164992
  https://developers.google.com/web/updates/2016/08/chrome-53-deprecations?hl=en#dhe-based-ciphers-being-phased-out


baloo:
see https://codereview.chromium.org/2056343006/diff/100001/net/socket/ssl_client_socket_impl.cc

jvehent:
This is not the right fix for your issue. 

> any previous cipher in the list doesn't match with the chrome internal cipher list

What version of openssl does your server use that it can't provide ECDHE ciphersuites before of DHE ?

baloo:
@jvehent You may reproduce with the beta release of chrome (https://www.google.com/chrome/browser/beta.html)  and by stripping all ECDHE from the cipher list. This is especially a problem on older releases of openssl with no support for ECDHE (eg: 0.9.8).

jvehent:
This is not how ciphersuite selection works. If Chrome 53 stops supporting DHE (it should), and a given server doesn't support ECDHE, the connection should fall back to a non PFS ciphersuite, like AES128-SHA.

baloo:
I agree with you but that not how it's implemented in chrome. Chrome will just trow the ERR_SSL_OBSOLETE_CIPHER if it can't negociate with higher priority ciphers and won't try the lower ones. 
Again, please try with the beta release :)

april:
@baloo: Can you provide me an example of an OpenSSL cipher list (and related protocol list) that fails in this way with Canary?  I'd be happy to test it out.  It seems extremely unlikely that Google would break classic RSA key exchange.

jvehent:
At any rate, based on the reading of Chrome plans, refusing to fall back to RSA key exchange would be a bug on their side.

This patch, while highlighting an interesting issue, is not something we want to merge.

baloo:
Well, sorry for the noise. I've just been bitten by our puppet which replaced my config file while I edited it. I can't reproduce. It's all good.

jvehent:
No need to apologize. This actually started an interesting discussion internally. I had no idea Chrome was about to remove DHE support entirely.

jrchamp:
It sounds like Chrome falls back to other ciphers now? In case it doesn't, here's a test case (all TLSv1.2):
openssl ciphers -v 'DHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-SHA256:DHE-RSA-AES128-SHA256:AES256-GCM-SHA384:AES128-GCM-SHA256:AES256-SHA256:AES128-SHA256'

Issue #156, opened 2016-08-24
Sweet32 Birthday attacks on 64-bit block ciphers in TLS


rugk:
See https://sweet32.info/

CVE-2016-2183, CVE-2016-6329

So what do you consider to do? Dropping 3DES from the Intermediate level?

And what to do about the [RC4 issue](https://wiki.mozilla.org/Security/Server_Side_TLS#RC4_weaknesses) now?

> In a public discussion ([bug 927045](https://bugzilla.mozilla.org/show_bug.cgi?id=927045)), it has been recommended to replace RC4 with 3DES. This would impact Internet Explorer 7 and 8 users that, depending on the OS, do not support AES, and will negotiate only RC4 or 3DES ciphers. Internet Explorer uses the cryptographic library “schannel”, which is OS dependent. schannel supports AES in Windows Vista, but not in Windows XP.
> 
> While 3DES provides more resistant cryptography, it is also 30 times slower and more cpu intensive than RC4. For large web infrastructure, the CPU cost of replacing RC4 with 3DES is non-zero. For this reason, we recommend that administrators evaluate their traffic patterns, and make the decision of replacing RC4 with 3DES on a per-case basis. At Mozilla, we evaluated that the impact on CPU usage is minor, and thus decided to replace RC4 with 3DES where backward compatibility is required.

I assume this attack makes the decision between RC4 and 3DES harder.

tomato42:
Exploiting the 3DES issue is harder in practice, as you can't keep connection open indefinitely against Apache or nginx, so 3DES is still better than RC4.

but I guess that means we should move Windows XP and Windows 2003 compatibility completely to legacy level.

jvehent:
I don't think we should rush removing 3DES from the intermediate configuration. That is certainly a long term goal, but [stats from the real world](https://jve.linuxwall.info/blog/index.php?post/2016/08/04/TLS-stats-from-1.6-billion-connections-to-mozilla.org) show that it is still significantly used (2.8%).

I would propose to continue the work that has started on v5 of the guidelines, planned for early 2017, and as @tomato42 proposes, try to move 3DES to the old level then.

april:
Even with modern clients, it's still in use about 1% of the time, which is pretty high:

https://ipv.sx/telemetry/general-v2.html?channels=release&measure=SSL_SYMMETRIC_CIPHER_FULL&target=4&absolute=0&relative=1

jvehent:
[...]

PR #160, opened 2016-08-30
Update index.html

dennis-fernuni:
DES (3DES) removed from intermediate-config


jvehent:
Thanks for the patch. Like I said in #156, this is something we need to spend more time evaluating, because 3DES still account for a significant portion of legitimate traffic. It will be part of v5 of the guidelines that we'll release likely at the end of the year. Until then, 3DES will remain in the intermediate guidelines.

PR #165, opened 2016-09-30
v4.1: Clarify Logjam notes, Clarify risk of TLS Tickets

jvehent:
r? @ekr @marumari

### DIFF
--- a/Server_Side_TLS.mediawiki
+++ b/Server_Side_TLS.mediawiki
@@ -262,15 +262,12 @@ MBYCEQCHU6UNZoHMF6bPtj21Hn/bAgEC.....
 </source>
 
 == Pre-defined DHE groups ==
-In order to lower the burden of system administrators, several servers provide pre-computed DH groups. Unfortunately, the [https://weakdh.org/ logjam] report showed that it is very likely that a state-level adversary may have broken the most widely used 1024-bit DH group, Oakley group 2, standardized in [https://tools.ietf.org/html/rfc2409#section-6.2 rfc2409]].
+In order to lower the burden of system administrators, several servers provide pre-computed DH groups. Unfortunately, the [https://weakdh.org/ logjam] shows that pre-defined 1024 bits DHE is not safe, and administrators should either:
+* use a larger group, with a minimum size of 2048-bit, as recommended in the intermediate configuration ;
+* keep using a 1024-bit DH group if you need to (see [[#DHE_and_Java]]), but generate DH parameters yourself via the openssl dhparam 1024 command ;
+* disable DHE altogether, as recommended in the modern configuration, and rely on ECDHE for PFS if you don't support legacy clients lacking ECDHE support (see [[#DHE_and_ECDHE_support]]).
 
-For this reason, the use of this group is considered unsafe and you should either:
-* use a larger group, with a minimum size of 2048-bit, as recommended in the intermediate and modern configurations ;
-* keep using a 1024-bit DH group if you need to (see [[#DHE_and_Java]]), but move away from Oakley group 2 and use a custom DH group instead, generated via the openssl dhparam 1024 command ;
-* disable DHE altogether, relying on ECDHE for PFS if you don't support legacy clients lacking ECDHE support (see [[#DHE_and_ECDHE_support]]).
-
-It is currently assumed that standardized 2048 bits DH groups provide sufficient security to resist factorization attacks. However, the careful administrator should generate a random DH group instead of using a
-standardized one when setting up a new server, as advised by the [https://weakdh.org|logjam] authors.
+When using pre-defined DH groups, the IETF recommends implementing [RFC 7919 https://tools.ietf.org/html/rfc7919] and use either the ffdhe2048, ffdhe3072 or ffdhe4196 group. However, the careful administrator should generate a random DH group instead of using a standardized one when setting up a new server.
 
 == DHE and ECDHE support ==
 Most modern clients that support both ECDHE and DHE typically prefer the former, because ECDHE provides faster handshakes than DHE ([http://vincent.bernat.im/en/blog/2011-ssl-perfect-forward-secrecy.html], [http://nmav.gnutls.org/2011/12/price-to-pay-for-perfect-forward.html]).
@@ -512,9 +509,9 @@ SPDY version 3 is vulnerable to the CRIME attack (see also http://zoompf.com/201
 
 Once a TLS handshake has been negociated between the server and the client, both may exchange a session ticket, which contains an AES-CBC 128bit key which can decrypt the session. This key is generally static and only regenerated when the web server is restarted (with recent versions of Apache, it's stored in a file and also kept upon restarts).
 
-The current work-around is to disable RFC 5077 support.
+There is a security concern around the management of TLS tickets in web servers: if a server is breached, the key that encrypted tickets can be stolen and used to decrypted recorded TLS sessions. Administrators that are more concerned about security than performance may want to disable TLS tickets support entirely. However, this is a drastic measure that will impact performances, and thus the general recommendation is to implement regular restarts and wipes of the key file in web servers.
 
-more: https://media.blackhat.com/us-13/US-13-Daigniere-TLS-Secrets-Slides.pdf
+more information: https://media.blackhat.com/us-13/US-13-Daigniere-TLS-Secrets-Slides.pdf
 
 == Cipher names correspondence table ==
 IANA, OpenSSL and GnuTLS use different naming for the same ciphers. The table below matches these ciphers as well as their corresponding compatibility level.
@@ -2814,6 +2811,10 @@ In the example above, the component SIGN-RSA-SHA224 is not supported by this ver
 ! Version
 ! Editor
 ! Changes
+|-
+| style="text-align: center;" | 4.1
+| style="text-align: center;" | Julien Vehent
+| Clarify Logjam notes, Clarify risk of TLS Tickets
 |-
 | style="text-align: center;" | 4
 | style="text-align: center;" | Julien Vehent
--- a/Server_Side_TLS.mediawiki
+++ b/Server_Side_TLS.mediawiki
@@ -251,23 +251,56 @@ As an example, TLS_DHE_RSA_WITH_AES_128_CBC_SHA256 works as follow:
 # Client sends a [http://tools.ietf.org/html/rfc5246#section-7.1 CHANGE CIPHER SPEC] message to the server, and both parties continue the handshake using ENCRYPTED HANDSHAKE MESSAGES
 
 The size of the prime number ''p'' constrains the size of the pre-master key ''PMS'', because of the modulo operation. A smaller prime almost means weaker values of ''A'' and ''B'', which could leak the secret values ''X'' and ''Y''. Thus, the prime ''p'' should not be smaller than the size of the RSA private key.
-<source lang="bash">
-$ openssl dhparam 2048
-Generating DH parameters, 2048 bit long safe prime, generator 2
-..+..+...............+
+
+== Pre-defined DHE groups ==
+
+Instead of using pre-configured DH groups, or generating their own with "openssl dhparam", operators should use the pre-defined DH groups ffdhe2048, ffdhe3072 or ffdhe4096 recommended by the IETF in [RFC 7919 https://tools.ietf.org/html/rfc7919].  These groups are audited and may be more resistant to attacks than ones randomly generated.
+
+Note: if you must support older Java clients, Dh groups larger than 1024 bits may block connectivity (see [[#DHE_and_Java]]).
+
+=== ffdhe2048 ===
+<source>
 -----BEGIN DH PARAMETERS-----
-MBYCEQCHU6UNZoHMF6bPtj21Hn/bAgEC.....
-......
+MIIBCAKCAQEA//////////+t+FRYortKmq/cViAnPTzx2LnFg84tNpWp4TZBFGQz
++8yTnc4kmz75fS/jY2MMddj2gbICrsRhetPfHtXV/WVhJDP1H18GbtCFY2VVPe0a
+87VXE15/V8k1mE8McODmi3fipona8+/och3xWKE2rec1MKzKT0g6eXq8CrGCsyT7
+YdEIqUuyyOP7uWrat2DX9GgdT0Kj3jlN9K5W7edjcrsZCwenyO4KbXCeAvzhzffi
+7MA0BM0oNC9hkXL+nOmFg/+OTxIy7vKBg8P+OxtMb61zO7X8vC7CIAXFjvGDfRaD
+ssbzSibBsu/6iGtCOGEoXJf//////////wIBAg==
 -----END DH PARAMETERS-----
 </source>
 
-== Pre-defined DHE groups ==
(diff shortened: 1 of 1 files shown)


Server_Side_TLS.mediawiki
```diff
@@ -262,15 +262,12 @@ MBYCEQCHU6UNZoHMF6bPtj21Hn/bAgEC.....
 </source>
 
 == Pre-defined DHE groups ==
-In order to lower the burden of system administrators, several servers provide pre-computed DH groups. Unfortunately, the [https://weakdh.org/ logjam] report showed that it is very likely that a state-level adversary may have broken the most widely used 1024-bit DH group, Oakley group 2, standardized in [https://tools.ietf.org/html/rfc2409#section-6.2 rfc2409]].
+In order to lower the burden of system administrators, several servers provide pre-computed DH groups. Unfortunately, the [https://weakdh.org/ logjam] shows that pre-defined 1024 bits DHE is not safe, and administrators should either:
```
april: "the logjam shows" sounds weird.  I think it should probably be "the Logjam report suggests", capital L.

Server_Side_TLS.mediawiki
```diff
@@ -512,9 +509,9 @@ SPDY version 3 is vulnerable to the CRIME attack (see also http://zoompf.com/201
 
 Once a TLS handshake has been negociated between the server and the client, both may exchange a session ticket, which contains an AES-CBC 128bit key which can decrypt the session. This key is generally static and only regenerated when the web server is restarted (with recent versions of Apache, it's stored in a file and also kept upon restarts).
 
-The current work-around is to disable RFC 5077 support.
+There is a security concern around the management of TLS tickets in web servers: if a server is breached, the key that encrypted tickets can be stolen and used to decrypted recorded TLS sessions. Administrators that are more concerned about security than performance may want to disable TLS tickets support entirely. However, this is a drastic measure that will impact performances, and thus the general recommendation is to implement regular restarts and wipes of the key file in web servers.
```
april: Instead of wipes, perhaps "tickets with short lifetimes"?

Server_Side_TLS.mediawiki
ekr: I don't think this text is quite right either.

There are two forms of attack to be concerned with:
1. Passive
2. Active

It's true that there's a precomputation _passive_ attack on 1024-bit DHE but that's not shown by logjam; it's been known forever. What logjam shows is that there's an active attack even if the pairs prefer a stronger group, but then it's not clear how much the precomputation buys you because active attacks are generally targeted and if you can do the precomputation attack on 1024-bit DHE at all, then you could at least potentially do it for a few high-value targets (the ones you are likely to mount active attacks on).

Server_Side_TLS.mediawiki
```diff
@@ -262,15 +262,12 @@ MBYCEQCHU6UNZoHMF6bPtj21Hn/bAgEC.....
 </source>
 
 == Pre-defined DHE groups ==
-In order to lower the burden of system administrators, several servers provide pre-computed DH groups. Unfortunately, the [https://weakdh.org/ logjam] report showed that it is very likely that a state-level adversary may have broken the most widely used 1024-bit DH group, Oakley group 2, standardized in [https://tools.ietf.org/html/rfc2409#section-6.2 rfc2409]].
+In order to lower the burden of system administrators, several servers provide pre-computed DH groups. Unfortunately, the [https://weakdh.org/ logjam] shows that pre-defined 1024 bits DHE is not safe, and administrators should either:
+* use a larger group, with a minimum size of 2048-bit, as recommended in the intermediate configuration ;
```
ekr: with the impact that this messes up Java.

Server_Side_TLS.mediawiki
```diff
@@ -262,15 +262,12 @@ MBYCEQCHU6UNZoHMF6bPtj21Hn/bAgEC.....
...
+* keep using a 1024-bit DH group if you need to (see [[#DHE_and_Java]]), but generate DH parameters yourself via the openssl dhparam 1024 command ;
```
ekr: I've been trying to decide if this is a good recommendation. Would those people be better off with PFS @1024 versus non-PFS @2048?

Server_Side_TLS.mediawiki
```diff
@@ -262,15 +262,12 @@ MBYCEQCHU6UNZoHMF6bPtj21Hn/bAgEC.....
...
 
-For this reason, the use of this group is considered unsafe and you should either:
-* use a larger group, with a minimum size of 2048-bit, as recommended in the intermediate and modern configurations ;
-* keep using a 1024-bit DH group if you need to (see [[#DHE_and_Java]]), but move away from Oakley group 2 and use a custom DH group instead, generated via the openssl dhparam 1024 command ;
-* disable DHE altogether, relying on ECDHE for PFS if you don't support legacy clients lacking ECDHE support (see [[#DHE_and_ECDHE_support]]).
-
-It is currently assumed that standardized 2048 bits DH groups provide sufficient security to resist factorization attacks. However, the careful administrator should generate a random DH group instead of using a
-standardized one when setting up a new server, as advised by the [https://weakdh.org|logjam] authors.
+When using pre-defined DH groups, the IETF recommends implementing [RFC 7919 https://tools.ietf.org/html/rfc7919] and use either the ffdhe2048, ffdhe3072 or ffdhe4196 group. However, the careful administrator should generate a random DH group instead of using a standardized one when setting up a new server.
```
ekr: 1. The IETF does recommend 7919 but you can use the 7919 groups on your server regardless. This is what NSS does.
2. This second recommendation is, as I said, contrary to what IETF is currently recommending.

Server_Side_TLS.mediawiki
ekr: Tickets with short lifetimes does not help unless you also change keys. The right fix is to change keys frequently _and_ hint correctly to the client about the lifetime of the keys

Server_Side_TLS.mediawiki
jvehent: When we previously discussed this, the consensus was the risk of leaking a private key was likely greater than the risk of breaking through 1024 bits DHE. I don't have any metric to prove this though, only experience.

Server_Side_TLS.mediawiki
april: Do you have a link to 1. & 2. to source it?

Server_Side_TLS.mediawiki
jrchamp: "the key that encrypted tickets can be stolen" => "the key that encrypted tickets **use** can be stolen"
"used to decrypted recorded TLS sessions" => "used to **decrypt** recorded TLS sessions"
"TLS tickets support" => "TLS **ticket** support"
"will impact performances" => "will impact **performance**"

Server_Side_TLS.mediawiki
ekr: I'm not sure which claim you are requesting sources for. The RFC 7919 link is in ulfr's text. I think it's obvious that you can use 7919 in groups in your server, and NSS probably doesn't have a big enough penetration to make it worthwhile covering that.

As far as the second recommendation, I think you can take the fact that IETF published 7919 and that TLS 1.3 only supports predefined groups as evidence of that.

Server_Side_TLS.mediawiki
jvehent: @ekr: `hint correctly to the client` Could you clarify how this should be implemented in practice?

jrchamp (changes requested):
Thank you for clarifying this information, just a few small wording changes from me (in bold to highlight the words changed).

Server_Side_TLS.mediawiki
ekr: There's a protocol parameter field. I don't know whether OpenSSL lets you set it.

Server_Side_TLS.mediawiki
martinthomson: The reason for preconfigured groups isn't to reduce the burden on administrators, the problem with randomly generated groups is the computational load and that it is very difficult to validate a group for correctness. On a relatively fast machine generating a new group at even as small as 1024, it takes a noticeable amount of time.

Server_Side_TLS.mediawiki
```diff
@@ -512,9 +509,9 @@ SPDY version 3 is vulnerable to the CRIME attack (see also http://zoompf.com/201
 
 Once a TLS handshake has been negociated between the server and the client, both may exchange a session ticket, which contains an AES-CBC 128bit key which can decrypt the session. This key is generally static and only regenerated when the web server is restarted (with recent versions of Apache, it's stored in a file and also kept upon restarts).
```
martinthomson: Nit: negotiated

Server_Side_TLS.mediawiki
martinthomson: Framing this as a performance /security tradeoff might be right, but I think that the emphasis is wrong. The point you should be making is that managing a session ticket decryption key is difficult : you need to roll it frequently and a breach exposes multiple connections to attack.

s/performances/performance

leonklingele: `is to implement regular restarts and wipes of the key file in web servers`
At least with nginx it's sufficient to only reload (instead of restart) the web server to regenerate the session ticket key.

leonklingele: `and thus the general recommendation is to implement regular restarts and wipes of the key file in web servers.`
IMO it's better to completely disable session tickets if you don't need them (at least until there is a proper key rotation mechanism).

jvehent: I'll clarify further around Martin's comment. I think it's important to remember the target audience is operators who do not have control over the source code ran by their web servers, therefore our advice should be as generic as possible (eg. restart servers often).

Server_Side_TLS.mediawiki
jvehent: Would you say it's unreasonable advice to recommend operators to run `openssl dhparam 2048`, versus using preconfigured groups? The downside being web servers don't allow operators to configure a given group, so if dhparam isn't set by the ops, there's a risk of using a weak group.

I think the computational load is irrelevant here because only done once when configuring the service, or at regular interval, but definitely not during the handshake.

Server_Side_TLS.mediawiki
jvehent: Is this wording better? `If the web server does use one of these groups, and does not allow the administrator to configure them, the alternative is to generate a random DH group with "openssl dhparam" when setting up the server.`

jvehent: Nevermind, @ekr proposed to simply tell ops to use a PEM version of ffdhe_2048.

jvehent:
If you'd like to give this another read, I think it's ready to go. No more `openssl dhparam` though.

Server_Side_TLS.mediawiki
```diff
@@ -510,11 +539,11 @@ SPDY version 3 is vulnerable to the CRIME attack (see also http://zoompf.com/201
 
 == TLS tickets (RFC 5077) ==
 
-Once a TLS handshake has been negociated between the server and the client, both may exchange a session ticket, which contains an AES-CBC 128bit key which can decrypt the session. This key is generally static and only regenerated when the web server is restarted (with recent versions of Apache, it's stored in a file and also kept upon restarts).
+Once a TLS handshake has been negotiated between the server and the client, both may exchange a session ticket, which contains an AES-CBC 128bit key which can decrypt the session. This key is generally static and only regenerated when the web server is restarted (with recent versions of Apache, it's stored in a file and also kept upon restarts).
```
tomato42: The session ticket is _encrypted_ with a key, it doesn't contain a key for specific algorithm. Also, it only _usually_ is encrypted  with a AES-128-CBC.

Server_Side_TLS.mediawiki
```diff
@@ -510,11 +539,11 @@ SPDY version 3 is vulnerable to the CRIME attack (see also http://zoompf.com/201
...
 
-The current work-around is to disable RFC 5077 support.
+The key that encrypt TLS tickets in servers is very hard to manage and potentially induces a security risk if not renewed regularly: if a server is breached, the key can be stolen and used to decrypt recorded TLS tickets, thus leaking session keys. TLS tickets do bring a performance benefit because of session resumption, but administrators that are more concerned about security than performance may want to disable them entirely. The trade-off we recommend is to implement restarts of web servers and force deletion of local caches to renew encryption keys.
```
tomato42: Session resumption using session IDs also works quite well, if you don't use multiple servers that handle the requests.

Server_Side_TLS.mediawiki
```diff
@@ -251,26 +251,56 @@ As an example, TLS_DHE_RSA_WITH_AES_128_CBC_SHA256 works as follow:
...
-<source lang="bash">
-$ openssl dhparam 2048
-Generating DH parameters, 2048 bit long safe prime, generator 2
-..+..+...............+
+
+== Pre-defined DHE groups ==
+
+Instead of using pre-configured DH groups, or generating their own with "openssl dhparam", operators should use the pre-defined DH groups ffdhe2048, ffdhe3072 or ffdhe4096 recommended by the IETF in [RFC 7919 https://tools.ietf.org/html/rfc7919].  These groups are audited and may be more resistant to attacks than ones randomly generated.
+
+Note: if you must support older Java clients, Dh groups larger than 1024 bits may block connectivity (see [[#DHE_and_Java]]).
```
tomato42: given that JRE 6 is only supported with a paid-for contract with Oracle, I think we can just call them "old", not "older" Java clients. Also the H in "Dh" should be upper case, it's an acronym.

Server_Side_TLS.mediawiki
```diff
@@ -251,26 +251,56 @@ As an example, TLS_DHE_RSA_WITH_AES_128_CBC_SHA256 works as follow:
...
 
 The size of the prime number ''p'' constrains the size of the pre-master key ''PMS'', because of the modulo operation. A smaller prime almost means weaker values of ''A'' and ''B'', which could leak the secret values ''X'' and ''Y''. Thus, the prime ''p'' should not be smaller than the size of the RSA private key.
-<source lang="bash">
-$ openssl dhparam 2048
-Generating DH parameters, 2048 bit long safe prime, generator 2
-..+..+...............+
+
+== Pre-defined DHE groups ==
+
+Instead of using pre-configured DH groups, or generating their own with "openssl dhparam", operators should use the pre-defined DH groups ffdhe2048, ffdhe3072 or ffdhe4096 recommended by the IETF in [RFC 7919 https://tools.ietf.org/html/rfc7919].  These groups are audited and may be more resistant to attacks than ones randomly generated.
```
tomato42, line 257: > Instead of using pre-configured DH groups

I think it may be confusing about who pre-configured those DH groups. If it was previous administrator that generated a secure prime of 2048 bit length, it's suboptimal, but not catastrophic, unlike the 1024 bit Apache params.

tomato42 (changes requested):

mnordhoff:
The "DHE and Java" section still suggests "using custom 1024-bit DH parameters", but the document no longer says _how_, besides a passing reference to `openssl dhparam` in the "Pre-defined DHE groups" section. Should a brief `openssl dhparam 1024` example be added to the Java section, or should it be left as an exercise to the reader to discourage accommodating obsolete Java releases? Or should the suggestion be removed entirely?

(Hi, I'm Peng. I just wandered in.)

Server_Side_TLS.mediawiki
jrchamp: A couple nits: "The key that **encrypts** TLS tickets in servers is very hard to manage and potentially **introduces** a security risk if not renewed regularly:" or "The key **used to** encrypt TLS tickets in servers is very hard to manage and potentially **introduces** a security risk if not renewed regularly:"

PR #169, opened 2016-10-29
Update Server_Side_TLS.mediawiki

MichaelPaoli:
followed but --> followed by - hopefully corrected what appeared to be an English misuse or typo


jvehent:
Looks good to me, thanks for the patch!

PR #170, opened 2016-11-17
Add HSTS headers to Nginx error pages

sethp-jive:
via http://nginx.org/en/docs/http/ngx_http_headers_module.html#add_header

```
Adds the specified field to a response header provided that the response code equals 200, 201, 204, 206, 301, 302, 303, 304, or 307. ...

If the always parameter is specified (1.7.5), the header field will be added regardless of the response code.
```

See also: http://serverfault.com/a/647552

### DIFF
--- a/ssl-config-generator/index.html
+++ b/ssl-config-generator/index.html
@@ -336,7 +336,7 @@ $SERVER["socket"] == ":443" {
                         data.hsts = '\n    # HSTS (ngx_http_headers_module is required) (15768000 seconds = 6 months)' + '\n' +
                             '    add_header Strict-Transport-Security "max-age=15768000"';
                         if (isSemVer(data.serverVersion, '>=1.7.5')) {
-                            data.hsts += ' always'
+                            data.hsts += ' always';
                         }
                         data.hsts += ';';
                     }
--- a/ssl-config-generator/index.html
+++ b/ssl-config-generator/index.html
@@ -334,7 +334,7 @@ $SERVER["socket"] == ":443" {
                     }
                     if (data.hstsEnabled == "true") {
                         data.hsts = '\n    # HSTS (ngx_http_headers_module is required) (15768000 seconds = 6 months)' + '\n' +
-                            '    add_header Strict-Transport-Security max-age=15768000;';
+                            '    add_header Strict-Transport-Security "max-age=15768000" always;';
                     }
                     if (isSemVer(data.serverVersion, ">=1.9.5")) {
                         data.listen = '    listen 443 ssl http2;\n' +
--- a/ssl-config-generator/index.html
+++ b/ssl-config-generator/index.html
@@ -334,7 +334,12 @@ $SERVER["socket"] == ":443" {
                     }
                     if (data.hstsEnabled == "true") {
                         data.hsts = '\n    # HSTS (ngx_http_headers_module is required) (15768000 seconds = 6 months)' + '\n' +
-                            '    add_header Strict-Transport-Security "max-age=15768000" always;';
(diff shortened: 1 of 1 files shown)


floatingatoll:
This patch should test isSemVer >= 1.7.5, as it isn't available in prior versions of Nginx.

ssl-config-generator/index.html
```diff
@@ -334,7 +334,7 @@
                     }
                     if (data.hstsEnabled == "true") {
                         data.hsts = '\n    # HSTS (ngx_http_headers_module is required) (15768000 seconds = 6 months)' + '\n' +
-                            '    add_header Strict-Transport-Security max-age=15768000;';
+                            '    add_header Strict-Transport-Security "max-age=15768000" always;';
```
floatingatoll: 'always' here is only valid in nginx >= 1.7.5; per the example a couple lines below, this needs to be wrapped in a `isSemVer(data.serverVersion, ">=1.7.5")` test to ensure compatibility.

sethp-jive:
Thanks for the review – good catch! I think I've added in the conditional now, despite some missteps with using my browser as a code code editor.

floatingatoll:
(I filed #171 as well - we forgot the `isSemVer` on the Apache side, too!)

sethp-jive:
@floatingatoll ping – I think I added the isSemVer check, is there anything else?

floatingatoll:
Excellent! I don't have merge authority here, but those who do will see
this and respond at some point. It's holiday season right now so we're
rather short-staffed in many respects; some delay may occur.

 - R.

ssl-config-generator/index.html
```diff
@@ -334,7 +334,11 @@
                     }
                     if (data.hstsEnabled == "true") {
                         data.hsts = '\n    # HSTS (ngx_http_headers_module is required) (15768000 seconds = 6 months)' + '\n' +
-                            '    add_header Strict-Transport-Security max-age=15768000;';
+                            '    add_header Strict-Transport-Security "max-age=15768000"';
```
jvehent, line 337: Why has the header string changed as part of this patch? The double quotes have been added and the trailing semicolon removed.

jrchamp: Looks like a fix to treat it as a string?

sethp-jive: The final semicolon is added in line 341 below, but we have to do some logic before we're ready to terminate the directive (specifically, check if the server version if newer than 1.7.5).

The addition of the double quote is, I think, prophylactic here as `max-age=15768000` doesn't include any characters meaningful to the nginx parser. I'm not 100% sure since I haven't tried this config without the quotes, and all the examples I've seen delineate the header string this way. It's been a while since I stared deeply into the nginx config grammar, but I'm happy to chase that one down if you feel it's important.

sethp-jive:
hi @jrchamp @jvehent – is there any more information I can add here, or are we ready to merge?

ssl-config-generator/index.html
```diff
@@ -334,7 +334,11 @@
...
+                        if (isSemVer(data.serverVersion, '>=1.7.5')) {
+                            data.hsts += ' always'
```
jrchamp: It's not required, but might be nice to have a semicolon at the end of this statement.

jrchamp:
Everything makes sense to me. I noted a minor nit-pick thing, but nothing blocking from my view.

jrchamp (approved):
Love it. Header content is quoted and the "always" flag is added to the header command so that it outputs the header even on error pages.

april:
The new version of the generator fixes this and always includes `always` on the appropriate versions of nginx.  :)

PR #175, opened 2016-12-23
Add TLS 1.3, X25519, Disable DHE for Intermediate Level, Reordering Cipher Suites

zzq1015:
The original cipher suite is quite messy, therefore a cleanup is done.
Prioritizing ChaCha20 in modern config
OpenSSL 1.1.0 supports X25519 so it is added
ECDH curve order is changed to "X25519, secp256r1, secp521r1, secp384r1" because P-521 is slightly faster than P-384 in ECDH


jrchamp:
Side question: Does it make sense to undo the changes to server-side-tls-conf-4.0.json given that you've bumped the version and created server-side-tls-conf-4.2.json?

zzq1015:
OK maybe leave server-side-tls-conf-4.0.json as it was, and just update server-side-tls-conf and create server-side-tls-conf-4.2.json

Server_Side_TLS.mediawiki
```diff
@@ -35,170 +35,167 @@ The ordering of a ciphersuite is very important because it decides which algorit
 == <span style="color:green;">'''Modern'''</span> compatibility ==
 For services that don't need backward compatibility, the parameters below provide a higher level of security. This configuration is compatible with Firefox 27, Chrome 30, IE 11 on Windows 7, Edge, Opera 17, Safari 9, Android 5.0, and Java 8.
 
-* Ciphersuites: '''ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-SHA384:ECDHE-RSA-AES256-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256'''
-* Versions: '''TLSv1.2'''
-* TLS curves: '''prime256v1, secp384r1, secp521r1'''
-* Certificate type: '''ECDSA'''
-* Certificate curve: '''prime256v1, secp384r1, secp521r1'''
+* Ciphersuites: '''ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-SHA384:ECDHE-RSA-AES256-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256'''
```
jvehent: AES256-GCM is faster than CHACHA20 for CPUs that supports AESNI (most cpus targeted at the modern level do).

Server_Side_TLS.mediawiki
```diff
@@ -35,170 +35,167 @@ The ordering of a ciphersuite is very important because it decides which algorit
...
+* Versions: '''TLSv1.3 (working in progress), TLSv1.2'''
```
jvehent, line 39: I don't think we should recommend experimental versions here. We waited for CHACHA20 to be standardized before adding it to the guidelines, the same applies to TLS1.3.

Server_Side_TLS.mediawiki
```diff
@@ -35,170 +35,167 @@ The ordering of a ciphersuite is very important because it decides which algorit
...
+          0xC0,0x24 - ECDHE-ECDSA-AES256-SHA384 TLSv1.2 Kx=ECDH     Au=ECDSA Enc=AES(256)  Mac=SHA384
+          0xC0,0x28 - ECDHE-RSA-AES256-SHA384 TLSv1.2 Kx=ECDH     Au=RSA  Enc=AES(256)  Mac=SHA384
+          0xC0,0x23 - ECDHE-ECDSA-AES128-SHA256 TLSv1.2 Kx=ECDH     Au=ECDSA Enc=AES(128)  Mac=SHA256
+          0xC0,0x27 - ECDHE-RSA-AES128-SHA256 TLSv1.2 Kx=ECDH     Au=RSA  Enc=AES(128)  Mac=SHA256
 </source>
 
 Rationale:
-* AES256-GCM is prioritized above its 128 bits variant, and ChaCha20 because we assume that most modern devices support AESNI instructions and thus benefit from fast and constant time AES. 
-* We recommend ECDSA certificates with P256 as other curves may not be supported everywhere. RSA signatures on ECDSA certificates are permitted because very few CAs sign with ECDSA at the moment.
+* ChaCha20 > AES_256_GCM > AES_128_GCM > AES_256_CBC > AES_128_CBC because AES_GCM is fragile ([https://eprint.iacr.org/2013/157.pdf 1]) and hard to implement safely. Also, ChaCha20 is not necessarily slower than AES_256_GCM while providing 256 bits of security.
```
jvehent: This paper applies to AES_GCM implemented in OpenSSL, when the modern level specifically targets AESNI implementations. Do you have evidence to suggest AES_GCM is fragile when implemented in cpus?

Server_Side_TLS.mediawiki
```diff
@@ -35,170 +35,167 @@ The ordering of a ciphersuite is very important because it decides which algorit
...
-* Certificate signature: '''sha256WithRSAEncryption'''
+* Ciphersuites: '''ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES256-SHA:ECDHE-RSA-AES256-SHA:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES256-SHA384:ECDHE-RSA-AES256-SHA384:DHE-RSA-AES128-SHA:DHE-RSA-AES256-SHA:DHE-RSA-AES128-SHA256:DHE-RSA-AES256-SHA256:AES128-GCM-SHA256:AES256-GCM-SHA384:AES128-SHA:AES256-SHA:AES128-SHA256:AES256-SHA256:ECDHE-ECDSA-DES-CBC3-SHA:ECDHE-RSA-DES-CBC3-SHA:EDH-RSA-DES-CBC3-SHA:DES-CBC3-SHA'''
+* Versions: '''TLSv1.3 (working in progress), TLSv1.2, TLSv1.1, TLSv1'''
+* ECDH curves: '''X25519 (with OpenSSL 1.1.0+), prime256v1, secp521r1, secp384r1'''
+* Certificate type: '''RSA and ECDSA in parallel if available, otherwise just RSA'''
+* Certificate (ECDSA) curve: '''prime256v1, secp384r1, secp521r1'''
+* Certificate signature: '''sha256WithRSAEncryption for RSA, and ecdsa-with-SHA256, ecdsa-with-SHA384, ecdsa-with-SHA512 for ECDSA'''
```
jvehent, line 77: We can't use ECDSA certs in the intermediate level because the goal is to support a very large population of clients.

Server_Side_TLS.mediawiki
```diff
@@ -35,170 +35,167 @@ The ordering of a ciphersuite is very important because it decides which algorit
...
+          0x00,0x99 - DHE-DSS-SEED-SHA        SSLv3 Kx=DH       Au=DSS  Enc=SEED(128) Mac=SHA1
+          0x00,0x13 - DHE-DSS-DES-CBC3-SHA    SSLv3 Kx=DH       Au=DSS  Enc=3DES(168) Mac=SHA1
 </source>
 
 Rationale:
 * You should take a hard look at your infrastructure needs before using this configuration; it is intended for special use cases only, and most servers should use the intermediate configuration instead.
-* SSLv3 is enabled to support WinXP SP2 clients on IE.
-* SHA1 certificates are authorized but only via certificate switching, meaning the server must implement custom logic to provide a SHA1 certs to old clients, and SHA256 certs to all others. More information in the "Certificates Switching" section later in this document.
-* Most ciphers that are not clearly broken and dangerous to use are supported
+* SSLv3 is enabled to support IE6 on Windows XP.
```
jvehent, line 174: I believe Chrome on xpsp2 is also limited to ssl3, but I'd need to double check.

json/server-side-tls-conf-4.0.json
```diff
@@ -2,21 +2,21 @@
     "href": "https://statics.tls.security.mozilla.org/server-side-tls-conf-4.0.json",
```
jvehent: This file must remain unchanged for users who pin the version.

json/server-side-tls-conf.json
```diff
@@ -1,33 +1,33 @@
 {
-    "href": "https://statics.tls.security.mozilla.org/server-side-tls-conf.json",
+    "href": "https://statics.tls.security.mozilla.org/server-side-tls-conf-4.2.json",
```
jvehent, line 2: This URL must remain unchanged.

jvehent:
While I appreciate the patches, I think these changes are premature. TLS1.3 is still months away (and we're closely monitoring its progress), and adjustments to existing levels need further discussions.

I'd prefer to go back to the whiteboard, and discuss in an issue what recommendations we want to see in version 5.0 of the guidelines.

Server_Side_TLS.mediawiki
zzq1015: ChaCha20 is already standardized by IETF:
https://tools.ietf.org/html/rfc7539

Server_Side_TLS.mediawiki
zzq1015: I mean ECDSA and RSA **At The Same Time**.
Currently Nginx and Apache both have support for dual-certs environment
Apache: https://blog.joelj.org/dual-rsaecdsa-certificates-in-apache-2-4/
Nginx: https://nginx.org/en/CHANGES
```
Changes with nginx 1.11.0:
*) Feature: the "ssl_certificate" and "ssl_certificate_key" directives
can be specified multiple times to load certificates of different
types (for example, RSA and ECDSA).
```

Server_Side_TLS.mediawiki
zzq1015: I disagree. 

1. IE**6** on Windows XP (Any Service Pack) only enables SSLv2 and v3 by default, and TLSv1.0 can be enabled manually
2. Chrome is limited to RSA certificates on XP in **any service pack**, and XP pre-SP3 does not support SHA-2 (SHA224/256/384/512) certificates.
3. Chrome has deprecated SSLv3 long time ago, and nothing stops Chrome using TLSv1.2.

Server_Side_TLS.mediawiki
zzq1015: Sorry I made a mistake when doing OpenSSL speed test.
In OpenSSL 1.1.0 I did: 
`/path/to/openssl speed -evp chacha20`
instead of
`/path/to/openssl speed -evp chacha20-poly1305`
So my result for chacha20 is faster than aes-256-gcm but slower than aes-128-gcm.
chacha20-poly1305 is slower than both.
Speed test result:
```
type             16 bytes     64 bytes    256 bytes   1024 bytes   8192 bytes  16384 bytes
aes-128-gcm     215564.66k   516407.32k  1090354.18k  1310255.45k  1575836.68k  1826177.02k
aes-256-gcm     172686.72k   471236.02k   948903.34k  1151707.48k  1350737.92k  1275957.43k
chacha20        169704.14k   289461.82k   669630.89k  1384385.54k  1431011.33k  1535246.34k
chacha20-poly1305   131241.25k   237345.05k   473652.14k   882510.85k   881535.66k   924150.44k
```

Server_Side_TLS.mediawiki
zzq1015: Oh yeah hardware AES_GCM implementations are pretty robust, only the software implementations are fragile.

Server_Side_TLS.mediawiki
tomato42: are you sure that your OpenSSL 1.1.0 is compiled with AES-NI support and is running on CPU that supports it? I'm getting over 5GiB/s for aes-128-gcm on i7-6600U CPU @ 2.60GHz:
```
type             16 bytes     64 bytes    256 bytes   1024 bytes   8192 bytes  16384 bytes
aes-128-gcm     496908.87k  1272636.25k  2384261.14k  3661835.68k  5025197.46k  5093599.29k
aes-256-gcm     506347.22k  1202380.63k  2100516.94k  3170329.34k  3496019.56k  3559158.24k
chacha20        304854.80k   550133.96k  1324290.14k  2736866.65k  2819459.00k  2773707.09k
chacha20-poly1305   235724.31k   510298.69k   959805.56k  1741361.66k  1736591.29k  1791584.90k
```

Server_Side_TLS.mediawiki
```diff
@@ -35,170 +35,167 @@ The ordering of a ciphersuite is very important because it decides which algorit
...
 </source>
 
 Rationale:
-* AES256-GCM is prioritized above its 128 bits variant, and ChaCha20 because we assume that most modern devices support AESNI instructions and thus benefit from fast and constant time AES. 
-* We recommend ECDSA certificates with P256 as other curves may not be supported everywhere. RSA signatures on ECDSA certificates are permitted because very few CAs sign with ECDSA at the moment.
+* ChaCha20 > AES_256_GCM > AES_128_GCM > AES_256_CBC > AES_128_CBC because AES_GCM is fragile ([https://eprint.iacr.org/2013/157.pdf 1]) and hard to implement safely. Also, ChaCha20 is not necessarily slower than AES_256_GCM while providing 256 bits of security.
+* We recommend ECDSA certificates with NIST-P256 as other curves may not be supported everywhere. RSA signatures on ECDSA certificates are permitted because very few CAs sign with ECDSA at the moment.
 * DHE is removed entirely because it is slow in comparison with ECDHE, and all modern clients support elliptic curve key exchanges.
-* SHA1 signature algorithm is removed in favor of SHA384 for AES256 and SHA256 for AES128.
+* HMAC-SHA1 is removed in favor of HMAC-SHA384 for AES256 and HMAC-SHA256 for AES128.
```
tomato42: why remove HMAC-SHA1? it provides 160 bit level of security, higher than most other acceptable mechanisms in the FUTURE

Server_Side_TLS.mediawiki
tomato42: +1 for explicitly stating that server must have ECDSA and RSA certificates or just RSA certificates

though for ECDSA certificates, only P-256 and P-384 is acceptable - schannel doesn't support P-521 curve.

Server_Side_TLS.mediawiki
tomato42: No, schannel on Windows XP support TLS1.0, it (or rather IE6) just doesn't enable it by default.

Server_Side_TLS.mediawiki
zzq1015: @tomato42 That's my VPS
My computer with i7-6700HQ gets about 5GB/s for aes-128-gcm

Server_Side_TLS.mediawiki
zzq1015: @tomato42 The rationale was wrong. What they actually do is to remove HMAC-SHA1.
SHA-1 only provides about 80 bits of security, by having an output size of 160 bits.
SHA-256 has 128 bits of security, while having an output size of 256 bits
So, the bits of security is half of the output size.

zzq1015: Also, due to the weird cipher suite support in [IE 11 in WP8.1](https://dev.ssllabs.com/ssltest/viewClient.html?name=IE&version=11&platform=Win%20Phone%208.1&key=65) I'm adding back HMAC-SHA1 for AES256.

draft1:
Nginx already support TLS 1.3 via BoringSSL and OpenSSL 1.1.1.

And as of August 2018, the final version of TLS 1.3 is now published (RFC 8446)
https://tools.ietf.org/html/rfc8446

Chrome 70 and Firefox 63 also support the Final TLS 1.3.

synapt:
Finalized OpenSSL TLS 1.3 ciphers are the following;
```
TLS_AES_128_GCM_SHA256
TLS_AES_256_GCM_SHA384
TLS_CHACHA20_POLY1305_SHA256
TLS_AES_128_CCM_SHA256
TLS_AES_128_CCM_8_SHA256
```

However I've deployed all mine (with order) as;
```
TLS_CHACHA20_POLY1305_SHA256
TLS_AES_128_GCM_SHA256
TLS_AES_256_GCM_SHA384
```

Though there looks to be a potential bug in OpenSSL 1.1.1 at the moment in that it is not 'preserving' server-defined order for TLS 1.3 ciphers, only <= 1.2.

lukastribus:
Why would you enable and prioritize CCM suites? I'm not a crypto guy, but I assume there is a reason CCM suites are disabled by default in both openssl and browsers.

Server order seems to work correctly in TLSv1.3 with the command line server:
`openssl s_server -port 1443 -serverpref`

and with the [Simple TLS Server](https://wiki.openssl.org/index.php/Simple_TLS_Server) when adding:
`SSL_CTX_set_options(ctx, SSL_OP_CIPHER_SERVER_PREFERENCE);`

However it does not work with TLSv1.3 in other applications (specifically I tested haproxy). So digging up the trigger for this bug requires more investigation.

Edit: actually it looks like ssllabs test is just broken.

synapt:
You're correct, I admittedly didn't mean to include that one and copied while a bit sleepy lol.

And yeah OpenSSL itself seems to work fine for it, but other things are not (including ssllabs' test), so there might have been a mechanism change in how TLS 1.3 does defined-order, I haven't given the spec a deep look though, just figured it was worth a mention though.

lukastribus:
OpenSSL is fine, ssllabs is broken, I just opened [TLSv1.3 bug: server-preferred order detection broken due to Chacha20](https://github.com/ssllabs/ssllabs-scan/issues/657).

april:
Closing this as it has all been covered by Server Side TLS 5.0. Thank you so much for your work on this!

PR #183, opened 2017-03-02
Server Side TLS v5.0

jvehent:
This is the result of the discussion from https://github.com/mozilla/server-side-tls/issues/178

It's not ready to merge yet, but summarizes the goals of this new version. Please discuss.

### DIFF
--- a/Server_Side_TLS.mediawiki
+++ b/Server_Side_TLS.mediawiki
@@ -41,6 +41,7 @@ For services that don't need backward compatibility, the parameters below provid
 * Certificate type: '''ECDSA (recommended) or RSA'''
 * Certificate curve: '''prime256v1, secp384r1'''
 * Certificate signature: '''ecdsa-with-SHA256, ecdsa-with-SHA384, ecdsa-with-SHA512, sha256WithRSAEncryption, sha384WithRSAEncryption, sha512WithRSAEncryption'''
+* Certificate transparency: '''required'''
 * Bits of security: '''>=128''' (RSA 3072 and above, P-256, P-384)
 * HSTS: '''max-age=34387200''' (13 months)
 * HPKP: '''max-age=34387200''' (13 months)
@@ -74,6 +75,7 @@ For services that don't need compatibility with legacy clients (mostly WinXP), b
 * Certificate type: '''ECDSA (recommended) or RSA'''
 * Certificate curve: ''''prime256v1, secp384r1'''
 * Certificate signature: '''ecdsa-with-SHA256, ecdsa-with-SHA384, ecdsa-with-SHA512, sha256WithRSAEncryption, sha384WithRSAEncryption, sha512WithRSAEncryption'''
+* Certificate transparency: '''optional'''
 * Bits of security: '''>=112''' (RSA 2048 and above, P-256, P-384)
 * DH Parameter size: '''2048''' (recommend using [#Pre-defined_DHE_groups Pre-defined DHE groups])
 * HSTS: '''max-age=15768000''' (6 months)
@@ -119,6 +121,7 @@ This is the old ciphersuite that works with all clients back to Windows XP/IE6. 
 * Certificate type: '''RSA'''
 * Certificate curve: ''''None'''
 * Certificate signature: '''sha256WithRSAEncryption'''
+* Certificate transparency: '''optional'''
 * Bits of security: '''>=112''' (RSA 2048 and above)
 * DH Parameter size: '''1024'''
 * HSTS: '''max-age=2589000''' (1 month) (optional)
--- a/Server_Side_TLS.mediawiki
+++ b/Server_Side_TLS.mediawiki
@@ -36,8 +36,8 @@ The ordering of a ciphersuite is very important because it decides which algorit
 For services that don't need backward compatibility, the parameters below provide a higher level of security. This configuration is compatible with Firefox 27, Chrome 30, IE 11 on Windows 7, Edge, Opera 17, Safari 9, Android 5.0, and Java 8.
 
 * Ciphersuites: '''ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-SHA384:ECDHE-RSA-AES256-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256'''
-* Versions: '''TLSv1.2'''
-* TLS curves: '''prime256v1, secp384r1, secp521r1'''
+* Versions: '''TLSv1.3, TLSv1.2'''
+* TLS curves: '''prime256v1, secp384r1, x25519, x448'''
 * Certificate type: '''ECDSA (recommended) or RSA'''
 * Certificate curve: '''prime256v1, secp384r1, secp521r1'''
 * Certificate signature: '''ecdsa-with-SHA256, ecdsa-with-SHA384, ecdsa-with-SHA512, sha256WithRSAEncryption, sha384WithRSAEncryption, sha512WithRSAEncryption'''
@@ -66,8 +66,6 @@ Rationale:
 For services that don't need compatibility with legacy clients (mostly WinXP), but still need to support a wide range of clients, this configuration is recommended. It is is compatible with Firefox 1, Chrome 1, IE 7, Opera 5 and Safari 1.
 
 * Ciphersuites: '''ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA:ECDHE-RSA-AES256-SHA:AES128-GCM-SHA256:AES256-GCM-SHA384:AES128-SHA256:AES256-SHA256:AES128-SHA:AES256-SHA:!DSS'''
-* Versions: '''TLSv1.2, TLSv1.1, TLSv1'''
-* TLS curves: '''prime256v1, secp384r1, secp521r1'''
 * Certificate type: '''RSA'''
 * Certificate curve: ''''None'''
 * Certificate signature: '''sha256WithRSAEncryption'''
@@ -76,6 +74,8 @@ For services that don't need compatibility with legacy clients (mostly WinXP), b
 * ECDH Parameter size: '''256'''
 * HSTS: '''max-age=15768000'''
 * Certificate switching: '''None'''
+* Versions: '''TLSv1.3, TLSv1.2, TLSv1.1, TLSv1'''
+* TLS curves: '''prime256v1, secp384r1, x25519, x448'''
 
 <source>
 0xCC,0x14  -  ECDHE-ECDSA-CHACHA20-POLY1305  TLSv1.2  Kx=ECDH  Au=ECDSA  Enc=ChaCha20(256)  Mac=AEAD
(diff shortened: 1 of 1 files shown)


Server_Side_TLS.mediawiki
```diff
@@ -36,16 +36,18 @@ The ordering of a ciphersuite is very important because it decides which algorit
...
-* Certificate type: '''ECDSA'''
-* Certificate curve: '''prime256v1, secp384r1, secp521r1'''
-* Certificate signature: '''sha256WithRSAEncryption, ecdsa-with-SHA256, ecdsa-with-SHA384, ecdsa-with-SHA512'''
-* RSA key size: '''2048''' (if not ecdsa)
-* DH Parameter size: '''None''' (disabled entirely)
-* ECDH Parameter size: '''256'''
-* HSTS: '''max-age=15768000'''
-* Certificate switching: '''None'''
+* Versions: '''TLSv1.3, TLSv1.2'''
+* TLS curves: '''prime256v1, secp384r1, x25519, x448'''
```
ekr, line 40: I'm not sure I would recommend x448, given that Firefox doesn't do it. @davidben, does Chrome?

Server_Side_TLS.mediawiki
```diff
@@ -36,16 +36,18 @@ The ordering of a ciphersuite is very important because it decides which algorit
...
+* Certificate type: '''ECDSA (recommended) or RSA'''
+* Certificate curve: '''prime256v1, secp384r1'''
+* Certificate signature: '''ecdsa-with-SHA256, ecdsa-with-SHA384, ecdsa-with-SHA512, sha256WithRSAEncryption, sha384WithRSAEncryption, sha512WithRSAEncryption'''
```
ekr, line 43: I would nail the curves to the hashes TLS 1.3 style

Server_Side_TLS.mediawiki
```diff
@@ -36,16 +36,18 @@ The ordering of a ciphersuite is very important because it decides which algorit
...
+* Certificate transparency: '''required'''
```
ekr, line 44: I'm not sure why you're requiring this, but in any case it's not specific enough to tell people what to do.

Server_Side_TLS.mediawiki
```diff
@@ -63,23 +65,25 @@ For services that don't need backward compatibility, the parameters below provid
...
+* Versions: '''TLSv1.3, TLSv1.2, TLSv1.1, TLSv1'''
+* TLS curves: '''prime256v1, secp384r1, x25519, x448'''
+* Certificate type: '''ECDSA (recommended) or RSA'''
+* Certificate curve: ''''prime256v1, secp384r1'''
+* Certificate signature: '''ecdsa-with-SHA256, ecdsa-with-SHA384, ecdsa-with-SHA512, sha256WithRSAEncryption, sha384WithRSAEncryption, sha512WithRSAEncryption'''
+* Certificate transparency: '''optional'''
+* Bits of security: '''>=112''' (RSA 2048 and above, P-256, P-384)
+* DH Parameter size: '''2048''' (recommend using [#Pre-defined_DHE_groups Pre-defined DHE groups])
+* HSTS: '''max-age=15768000''' (6 months)
+* HPKP: '''max-age=2589000''' (1 month)
```
ekr, line 83: What's your rationale for lowering these values.

Also see comments above which apply here.

Server_Side_TLS.mediawiki
```diff
@@ -63,23 +65,25 @@ For services that don't need backward compatibility, the parameters below provid
...
-* Certificate type: '''RSA'''
-* Certificate curve: ''''None'''
-* Certificate signature: '''sha256WithRSAEncryption'''
-* RSA key size: '''2048'''
-* DH Parameter size: '''2048'''
-* ECDH Parameter size: '''256'''
-* HSTS: '''max-age=15768000'''
-* Certificate switching: '''None'''
+* Ciphersuites: '''ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA:ECDHE-RSA-AES256-SHA:AES128-GCM-SHA256:AES256-GCM-SHA384:AES128-SHA256:AES256-SHA256:AES128-SHA:AES256-SHA:!DSS'''
+* Versions: '''TLSv1.3, TLSv1.2, TLSv1.1, TLSv1'''
```
ekr, line 74: Why needs TLSv1?

ekr (changes requested):

Server_Side_TLS.mediawiki
davidben: Nope.

(Is that intended to be a preference order? Putting RFC 7748 after NIST curves is odd.)

jrchamp: Agree on ordering, SafeCurves first please

jrchamp: Even if nothing supports it right now, x448 is our only standard-driven, SafeCurve option above x25519. If the server does not support it, it should just skip over it, but the sooner we start pushing for it, the sooner support will be available on both the server and client side. Given that, if this set of recommendations is not the right place for advocating what's next, then I would understanding delaying its inclusion.

Server_Side_TLS.mediawiki
```diff
@@ -63,23 +65,25 @@ For services that don't need backward compatibility, the parameters below provid
...
+* TLS curves: '''prime256v1, secp384r1, x25519, x448'''
```
jrchamp, line 75: The curve selection and ordering discussion in Modern applies here as well.

Server_Side_TLS.mediawiki
jrchamp: For what it's worth, if you are getting rid of TLSv1, consider dropping TLSv1.1 as well given there are very few clients/servers that have a max version of TLSv1.1.

Server_Side_TLS.mediawiki
tomato42: `x25519` should be preferred, but it also should be optional

Server_Side_TLS.mediawiki
tomato42: because close to 20% of android phones are still on version that does not enable/support TLS1.2 (that is Android earlier than 4.3): https://www.statista.com/statistics/271774/share-of-android-platforms-on-mobile-devices-with-android-os/

Server_Side_TLS.mediawiki
```diff
@@ -63,23 +65,25 @@ For services that don't need backward compatibility, the parameters below provid
...
+* Certificate type: '''ECDSA (recommended) or RSA'''
```
tomato42, line 76: for intermediate, I think we should recommend ECDSA and RSA - dual cert deployment - or RSA alone

Server_Side_TLS.mediawiki
tomato42: +1, don't see a reason to lower them in the intermediate

Server_Side_TLS.mediawiki
```diff
@@ -36,16 +36,18 @@ The ordering of a ciphersuite is very important because it decides which algorit
...
+* Certificate curve: '''prime256v1, secp384r1'''
+* Certificate signature: '''ecdsa-with-SHA256, ecdsa-with-SHA384, ecdsa-with-SHA512, sha256WithRSAEncryption, sha384WithRSAEncryption, sha512WithRSAEncryption'''
+* Certificate transparency: '''required'''
+* Bits of security: '''>=128''' (RSA 3072 and above, P-256, P-384)
+* HSTS: '''max-age=34387200''' (13 months)
+* HPKP: '''max-age=34387200''' (13 months)
+* Certificate Validity: '''<=398 days''' (13 months)
+* OCSP Stapling: '''required'''
+* CAA: '''recommended'''
```
tomato42, line 50: `CAA` is not explained in the document

Server_Side_TLS.mediawiki
tomato42: yes, an explanation what it is and how it's handled is needed

Server_Side_TLS.mediawiki
```diff
@@ -98,42 +100,38 @@ For services that don't need compatibility with legacy clients (mostly WinXP), b
...
 0x00,0x35  -  AES256-SHA                     SSLv3    Kx=RSA   Au=RSA    Enc=AES(256)       Mac=SHA1
-0x00,0x0A  -  DES-CBC3-SHA                   SSLv3    Kx=RSA   Au=RSA    Enc=3DES(168)      Mac=SHA1
 </source>
 
 Rationale:
 * ChaCha20 is prefered as the fastest and safest in-software cipher, followed by AES128. Unlike the modern configuration, we do not assume clients support AESNI and thus do not prioritize AES256 above 128 and ChaCha20. There has been discussions ([http://www.mail-archive.com/dev-tech-crypto@lists.mozilla.org/msg11247.html 1], [http://www.mail-archive.com/dev-tech-crypto@lists.mozilla.org/msg12398.html 2]) on whether AES256 extra security was worth its computing cost in software (without AESNI), and the results are far from obvious. At the moment, AES128 is preferred, because it provides good security, is really fast, and seems to be more resistant to timing attacks.
-* DES-CBC3-SHA and EDH-RSA-DES-CBC3-SHA are maintained for backward compatibility with clients that do not support AES.
+* DHE is removed entirely because it is slow in comparison with ECDHE, and all modern clients support elliptic curve key exchanges.
```
tomato42, line 113: As I said in  https://github.com/mozilla/server-side-tls/issues/178#issuecomment-285023081 use of DHE can be valid.

tomato42 (changes requested):

Server_Side_TLS.mediawiki
tomato42: I wonder if we shouldn't dedicate at least a paragraph to RSA-PSS signatures and certificates...

Server_Side_TLS.mediawiki
```diff
@@ -63,23 +65,25 @@ For services that don't need backward compatibility, the parameters below provid
...
-* Certificate switching: '''None'''
+* Ciphersuites: '''ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA:ECDHE-RSA-AES256-SHA:AES128-GCM-SHA256:AES256-GCM-SHA384:AES128-SHA256:AES256-SHA256:AES128-SHA:AES256-SHA:!DSS'''
+* Versions: '''TLSv1.3, TLSv1.2, TLSv1.1, TLSv1'''
+* TLS curves: '''prime256v1, secp384r1, x25519, x448'''
+* Certificate type: '''ECDSA (recommended) or RSA'''
+* Certificate curve: ''''prime256v1, secp384r1'''
+* Certificate signature: '''ecdsa-with-SHA256, ecdsa-with-SHA384, ecdsa-with-SHA512, sha256WithRSAEncryption, sha384WithRSAEncryption, sha512WithRSAEncryption'''
+* Certificate transparency: '''optional'''
+* Bits of security: '''>=112''' (RSA 2048 and above, P-256, P-384)
+* DH Parameter size: '''2048''' (recommend using [#Pre-defined_DHE_groups Pre-defined DHE groups])
```
ghost, line 81: Do we still need to use DH parameters, even though we don't use DHE in the intermediate level anymore?

Server_Side_TLS.mediawiki
martinthomson: We shouldn't be recommending HPKP.

Server_Side_TLS.mediawiki
martinthomson: I think that x25519, p256, and p384 are fine here.  x448 might be awesome in theory, but it's not awesome in practice and we shouldn't tell people to put the effort in for things that they won't get.

Server_Side_TLS.mediawiki
martinthomson: These here are right; you want these in the certificate, and I think that we can rely on TLS stacks to get the PSS thing right.  While text on PSS might help, it's probably not going to get the results you want (we might see PSS certificates, which won't work).

martinthomson (commented):
Time to think about shipping this.  TLS 1.3 is going to RFC very soon.

jrchamp:
Is it too soon to mention [EdDSA](https://tools.ietf.org/html/rfc8410)? It seems like a very good option, for software that provides it as an option.

ekr:
Does any common CA issue EdDSA certs?

clintwilson:
Not for public trust certs at least; Edwards curves aren't approved in the BRs yet.

ekr:
In that case, yes, I think it's premature/

Server_Side_TLS.mediawiki
tomato42: > we might see PSS certificates, which won't work

they do work in OpenSSL and GnuTLS already; NSS is also slowly progressing towards that direction

Server_Side_TLS.mediawiki
ekr: I agree with @martinthomson, the world is moving away from HPKP

jvehent: Ack. I'll remove it altogether.

tomato42:
now it's about 4.4% of Android phones that can't connect to TLS 1.2 servers, and with the recent discussions on IETF TLS WG ML, we may actually be able to push for TLS 1.1 and TLS 1.0 deprecation (allow them only in "Old")

april:
Mind if I close this in favor of #255?

april:
Closing due to obsolescence.

PR #195, opened 2017-07-13
Update CHACHA20 cipher ids

### LINKED ISSUE #155: Chacha cipher IDs are not from published RFC
The numerical IDs at the https://wiki.mozilla.org/Security/Server_Side_TLS use the numbers from agl draft, not the published RFC (https://tools.ietf.org/html/rfc7905).

The RFC uses the following IDs:

```
TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256   = {0xCC, 0xA8}
TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256 = {0xCC, 0xA9}
```

while the document uses those IDs:

```
0xCC,0x14  -  ECDHE-ECDSA-CHACHA20-POLY1305  TLSv1.2  Kx=ECDH  Au=ECDSA  Enc=ChaCha20(256)  Mac=AEAD
0xCC,0x13  -  ECDHE-RSA-CHACHA20-POLY1305    TLSv1.2  Kx=ECDH  Au=RSA    Enc=ChaCha20(256)  Mac=AEAD
```

liggitt:
Ids taken from https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml

Fixes #155

### DIFF
--- a/Server_Side_TLS.mediawiki
+++ b/Server_Side_TLS.mediawiki
@@ -32,6 +32,8 @@ Older versions of OpenSSL may not return the full list of algorithms. AES-GCM an
 
 The ordering of a ciphersuite is very important because it decides which algorithms are going to be selected in priority. Each level shows the list of algorithms returned by its ciphersuite. If you have to pick ciphers manually for your application, make sure you keep the ordering.
 
+The ciphersuite numbers listed come from the IANA [https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-4 TLS Cipher Suite Registry]. Previous versions of these recommendations included draft numbers for ECDHE-ECDSA-CHACHA20-POLY1305 (0xCC,0x14) and ECDHE-RSA-CHACHA20-POLY1305 (0xCC,0x13).
+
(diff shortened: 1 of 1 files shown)


tomato42:
I think the old ones should be left in, but marked as old (because they are widely deployed)

liggitt:
I could see a comment, but I don't think the non-standard ones should remain as distinct rows

lukastribus:
Where are exactly are the draft ciphers widely deployed?

Google services don't support them, and other than older boringssl builds I don't think any ssl lib does. OpenSSL certainly never did.

So unless there is evidence about widespread adoption, I suggest we drop 'em. Its complicated enough already, no need to further confuse everyone with obsolete pre-standard ciphers ids.

jvehent:
Cloudflare deploys then, and cloudflare is a wide chunk of the internet, thus the ciphers are widely deployed.

That said, I don't disagree about dropping them entirely.

liggitt:
updated with a note, PTAL

liggitt:
@jvehent any other feedback?

tomato42 (approved):

PR #202, opened 2017-10-15
Update released versions

Sp1l:
Add new versions for
 * apache (2.2 and 2.4 patch versions)
 * nginx (patch versions)
 * lighttpd 
 * haproxy (1.6 and 1.7 branch missing)
 * openssl (1.1.0 is not a release, many patch-versions missing)

Not sure how useful it is to have all these versions in the list.
Still need to figure out if there are optimizations for HAProxy 1.6 and 1.7 branches


gstrauss:
FYI: lighttpd 1.4.46 was just released, and the versions list for lighttpd stops at 1.4.42.  The list is missing the lighttpd releases from the last year.

PR #206, opened 2018-02-27
Updated JavaScript Libraries and Added SRI Hashes

ghost:


gene1wood:
Thanks for the contribution! Tested it and looks good.

PR #210, opened 2018-03-17
Rename EDH to DHE

injust:
We should follow OpenSSL's cipher naming conventions for DHE ciphersuites.


MrKeiKun:
Waiting for this to get merged..

tomato42 (changes requested):
the EDH is supported by both 1.0.x and 1.1.x series, DHE- is supported by only 1.1.x

injust (commented):
I fail to see how `DHE_` is unsupported in OpenSSL's 1.0.2 branch; the `DHE-RSA-DES-CBC3-SHA` ciphersuite works perfectly fine.
Also, the RFCs for SSL 3.0, TLS 1.0, TLS 1.1, and TLS 1.2 all specify ciphersuites using `DHE_`, not `EDH_`.

jrchamp:
Here's the error I get on OpenSSL 1.0.1l

```
# openssl ciphers -v 'DHE-RSA-DES-CBC3-SHA'
Error in cipher list
error:1410D0B9:SSL routines:SSL_CTX_set_cipher_list:no cipher match:ssl_lib.c:1292:
# openssl ciphers -v 'EDH-RSA-DES-CBC3-SHA'
EDH-RSA-DES-CBC3-SHA    SSLv3 Kx=DH       Au=RSA  Enc=3DES(168) Mac=SHA1
```

injust:
@jrchamp Anything prior to the 1.0.2 branch has been EOL'd for quite some time, I'm not sure which versions of OpenSSL this guide is targeting for support, but 1.0.2 is the current LTS branch.
I understand maintaining compatibility with older clients, but on the server side, 1.0.1 has long since been deprecated.

jrchamp:
@injust RHEL6 is still fairly popular https://access.redhat.com/articles/1462223

Ultimately, this change only affects a 3DES cipher. The cipher itself is legacy, so when we drop the cipher, it'll fix itself. https://csrc.nist.gov/news/2017/update-to-current-use-and-deprecation-of-tdea

injust:
@jrchamp Fair enough.

PR #213, opened 2018-04-19
Update default versions, httpd and openssl was EOL

NoNameForMee:
It should perhaps be made somewhat simpler for people who are not running software which is EOL. With this change, the work of having to manually change versions of apache httpd and openssl is forced upon the people who run EOL software instead of people with up to date software.

Previous default version of apache httpd (2.2.15) is EOL since 2018/01/01 and no longer maintained. Previous default version of openssl (1.0.1e) is also no longer maintained (changed openssl to 1.0.2 which is LTS).

Sources:
https://httpd.apache.org/
https://www.openssl.org/policies/releasestrat.html

### DIFF
--- a/ssl-config-generator/index.html
+++ b/ssl-config-generator/index.html
@@ -429,8 +429,8 @@ $SERVER["socket"] == ":443" {
             function loadFromQueryString() {
                 // http://stackoverflow.com/a/10834119/837015
                 var defaults = {
-                    "server": "apache-2.2.15",
-                    "openssl": "1.0.1e",
+                    "server": "apache-2.4.28",
+                    "openssl": "1.0.2l",
                     "hsts": "yes",
                     "profile": "intermediate"
                 };


jrchamp:
This change updates the defaults to httpd 2.4.28 (current 2.4 is 2.4.33) and openssl 1.0.2l (current 1.0.2 is 1.0.2o).

Really, we should probably stop separating the versions except as it has an effect on the results. That would help mitigate the changes needed as new versions are released that don't have an effective difference for the user.

The effect on this change is that we should pick the most recent version of each software that changes the configuration output and append something like "and newer" to the version.

floatingatoll:
Has RHEL6 ended their support for httpd 2.2 as well?

jrchamp:
RHEL 6.9 is still httpd 2.2 (see: https://access.redhat.com/solutions/445713 )

Side note from the same article: there may be Software Collections support for httpd 2.4 on RHEL 6

NoNameForMee:
The sole intention with this request was to make it easer for the people running the upstream supported versions (by changing the default versions of Apache/Openssl to the latest versions included here). Anyone running any sort of special OS, which may do backports and security fixes separately for the upstream versions of Apache and Openssl (or any other package for that matter), should still be able to find the appropriate settings for their system and software combination.

However, as there is no "Select your OS button and we will provide appropriate settings for its included version of Apache and/or Openssl" I still think it is best for the settings which are present in the generator (default Apache and Openssl) to correspond to the upstream officially supported versions (of Apache and Openssl).

The older versions are still here, but would require any user of the generator to change the values to their own older version. This work (however minor) should not mostly fall on people running the upstream supported versions.

april:
This is fixed in the newest version of the generator:

tomato42:
> The older versions are still here, but would require any user of the generator to change the values to their own older version. This work (however minor) should not mostly fall on people running the upstream supported versions.

it's very unlikely that any backported changes would break compatibility with upstream released version - if anything, some features that are supported may be disabled by using the old config file, but it shouldn't result in server less secure than the corresponding upstream version

PR #216, opened 2018-06-19
Remove discussion of SPDY. It has been superseded by HTTP/2.

ekr:


tomato42:
not sure if we should be quiet about SPDY then

I think the page should either say that the implementation needs to support HTTP/2 when it supports SPDY or it should just have SPDY disabled (as it is an experimental protocol)

april:
All of the non-essential discussion has been removed from the newest version of the guidelines, which covers this as well.

PR #220, opened 2018-09-07
add missing 3DES openssl ciphers in table

lukastribus:
Adds 2 missing 3DES openssl counterparts in the Cipher names correspondence table.


tomato42 (approved):

april:
The table is no longer included, but I do hope to get the table fixed and included elsewhere someday.  :)

PR #225, opened 2018-10-25
Add new versions of software

Sp1l:
Add additional versions for all lists of software


lukastribus:
Regarding haproxy, please you drop the -rc and -dev releases. Development tags don't belong here, I don't know why they have been added in the first place.

Also, the upcoming 1.8.15 release will support setting TLSv1.3 ciphersuites via [new ciphersuite directives](http://git.haproxy.org/?p=haproxy-1.8.git;a=commitdiff;h=4be76416751aa22992a44f2f5cfdba506809fd89), so we maybe already include it here.

Sp1l:
> please you drop the -rc and -dev releases. Development tags don't belong here

Fully agree. From the list I see on http://git.haproxy.org/?p=haproxy.git I assumed that all these point releases were non-rc, non-dev. I'll remove the added versions as I can't determine which is which.

lukastribus:
Ok, I will clean up the haproxy releases here when I post the PR for the ciphersuites directives.

april:
The new generator doesn't use software version lists:

Issue #214, opened 2018-05-08
Invalid ELB configuration generated


captn3m0:
The ELB configuration generated has some invalid cipher suites (which AWS does not accept):

<details>
<summary> Running: 

`aws elb describe-load-balancer-policies --policy-names ELBSecurityPolicy-2016-08`
</summary>

```json
{
    "PolicyDescriptions": [
        {
            "PolicyName": "ELBSecurityPolicy-2016-08",
            "PolicyTypeName": "SSLNegotiationPolicyType",
            "PolicyAttributeDescriptions": [
                {
                    "AttributeName": "Protocol-TLSv1",
                    "AttributeValue": "true"
                },
                {
                    "AttributeName": "Protocol-SSLv3",
                    "AttributeValue": "false"
                },
                {
                    "AttributeName": "Protocol-TLSv1.1",
                    "AttributeValue": "true"
                },
                {
                    "AttributeName": "Protocol-TLSv1.2",
                    "AttributeValue": "true"
                },
                {
                    "AttributeName": "Server-Defined-Cipher-Order",
                    "AttributeValue": "true"
                },
                {
                    "AttributeName": "ECDHE-ECDSA-AES128-GCM-SHA256",
                    "AttributeValue": "true"
                },
                {
                    "AttributeName": "ECDHE-RSA-AES128-GCM-SHA256",
                    "AttributeValue": "true"
                },
                {
                    "AttributeName": "ECDHE-ECDSA-AES128-SHA256",
                    "AttributeValue": "true"
                },
                {
                    "AttributeName": "ECDHE-RSA-AES128-SHA256",
                    "AttributeValue": "true"
                },
                {
                    "AttributeName": "ECDHE-ECDSA-AES128-SHA",
                    "AttributeValue": "true"
                },
                {
                    "AttributeName": "ECDHE-RSA-AES128-SHA",
                    "AttributeValue": "true"
                },
                {
                    "AttributeName": "DHE-RSA-AES128-SHA",
                    "AttributeValue": "false"
                },
                {
                    "AttributeName": "ECDHE-ECDSA-AES256-GCM-SHA384",
                    "AttributeValue": "true"
                },
                {
                    "AttributeName": "ECDHE-RSA-AES256-GCM-SHA384",
                    "AttributeValue": "true"
                },
                {
                    "AttributeName": "ECDHE-ECDSA-AES256-SHA384",
                    "AttributeValue": "true"
                },
                {
[...]

PR #227, opened 2018-12-21
Adds link to terraform module for AWS config

captn3m0:
This was created out of a need because even the default ELB Configurations are not easily available in Terraform: https://github.com/terraform-providers/terraform-provider-aws/issues/1669

It supports all 3 mozilla configurations, and they are generated from the JSON files in this repo. Working on this was what led to #214 and #215 incidentally.


april:
If you want to generate a template for this for the generator, I'd be happy to accept a PR here:
https://github.com/mozilla/ssl-config-generator

Thanks!

PR #229, opened 2019-02-10
Update web server version lists

KarenAvetyan:
Updated web server version list and added openssl 1.1.1 branch for future TLS 1.3 configuration


april:
The new generator doesn't use software versions in quite this way, so I'm closing this. But thanks!

PR #234, opened 2019-04-10
Replace links to CDN hosted resources to CloudFlare

### LINKED ISSUE #232: Feature Request: replace the Google CDN with other provider
This webpage is now using Google CDN (ajax.googleapis.com), but maybe it should be replaced, which may make this website more accessible.
### Reason:

- If these javascript are not loaded, the website won't work.
- There are many nice public CDNs, like cdnjs and jsdelivr, which have better performance.
- CDNs like jsdelivr integrate many different providers, and have more uptime for users.
- All these CDNs are available worldwide, but Google CDN doesn't. For example, the CDNs above are all available in mainland China and jsdelivr even have servers in China. But ajax.googleapis is not a legal website in China and not accessible, which makes 20% of people in the world unavailable to use this website.

gene1wood:
This seeks to improve the user privacy posture of the site
This also wraps the google analytics in a DNT check

Details on CloudFlare's privacy policy : https://www.cloudflare.com/privacypolicy/

Fixes #232

### DIFF
--- a/ssl-config-generator/index.html
+++ b/ssl-config-generator/index.html
@@ -6,7 +6,7 @@
     <title>Generate Mozilla Security Recommended Web Server Configuration Files</title>
     <link rel="stylesheet" href="../css/screen.css" />
     <link rel="stylesheet" href="https://mozorg.cdn.mozilla.net/media/css/tabzilla-min.css" />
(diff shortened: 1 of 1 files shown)


popcorner:
It looks good, and these scripts can be loaded via mainland China IP now.
The performance via Hong Kong IP and Japan IP are better than before.

april:
This is done in the new version:

gene1wood:
As the new version includes all the js and hosts them at github (instead of a CDN)

PR #235, opened 2019-04-10
Remove tabzilla

### LINKED ISSUE #233: Tabzilla in ssl config generator is not working
When visiting https://mozilla.github.io/server-side-tls/ssl-config-generator/ ,the file [tabzilla.js](https://mozorg.cdn.mozilla.net/en-US/tabzilla/tabzilla.js) can't be loaded and return 403 Forbidden.

gene1wood:
Tabzilla was deprectated in mozilla/tabzilla#87
Fixes #233


april:
This is done in the new version:

Issue #239, opened 2019-04-24
Each piece of software should have its own template


april:
That would let us break things cleanly and easily. It would also make it simple to add new software to the generator.

april:
Done!

PR #242, opened 2019-05-02
Brand new config generator

april:
Don't merge this yet, this is just for looking over the initial rewrite. Still lots of work to go.


generator/src/js/helpers/normalizever.js
```diff
@@ -0,0 +1,21 @@
...
+  version = String(version);
+
+  // get the number of subversions, e.g. 1.0.1 = 2
+  const subversions = version.match(/\./g) === null ? 0 : version.match(/\./g).length;
+
+  // normalize based on the number of subversions
+  if (subversions === 0) {
+    version = `${version}.0.0`;  // 1 --> 1.0.0
+  } else if (subversions === 1) {
+    version = `${version.split('.')[0]}.0.${version.split('.')[1]}`;  // 1.1e --> 1.0.1e
```
gene1wood: Huh, what's the rationale for this conversion?

generator/src/templates/partials/apache.hbs
```diff
@@ -0,0 +1,59 @@
+<VirtualHost *:443>
+    ...
+    SSLEngine on
+{{#if (minver "2.4.8" form.serverVersion)}}
```
gene1wood, line 4: May want to indent this like line 14 or de-indent 14 to match this

generator/src/templates/partials/apache.hbs
```diff
@@ -0,0 +1,59 @@
...
+    Header always set Strict-Transport-Security "max-age={{output.hstsMaxAge}}"
+    {{/if}}
+    ...
+</VirtualHost>
+
+# {{form.config}} configuration, tweak to your needs
+SSLProtocol             {{#if (minver "2.3.16" form.serverVersion)}}{{replace output.protocols " -SSLv2" ""}}{{else}}{{output.protocols}}{{/if}}
+SSLCipherSuite          {{output.cipherSuites}}
+SSLHonorCipherOrder     on
+{{#if (minver "0.9.8" form.opensslVersion)}}
```
gene1wood: Maybe through in some breadcrumbs to the docs that explain this complex conditional section
```
{{#if (minver "0.9.8" form.opensslVersion)}}
  // https://httpd.apache.org/docs/2.2/mod/mod_ssl.html#sslcompression
  {{#if (matchesminorver "2.2" form.serverVersion)}}
    {{#if (minver "2.2.24" form.serverVersion)}}
SSLCompression          off
    {{/if}}
  {{/if}}
  // https://httpd.apache.org/docs/2.4/mod/mod_ssl.html#sslcompression
  {{#if (matchesminorver "2.4" form.serverVersion)}}
    {{#if (minver "2.4.3" form.serverVersion)}}
SSLCompression          off
    {{/if}}
  {{/if}}
{{/if}}
{{#if (minver "0.9.8f" form.opensslVersion)}}
  // https://httpd.apache.org/docs/2.2/mod/mod_ssl.html#sslsessiontickets
  {{#if (matchesminorver "2.2" form.serverVersion)}}
    {{#if (minver "2.2.30" form.serverVersion)}}
SSLSessionTickets       off
    {{/if}}
  {{/if}}
  // https://httpd.apache.org/docs/2.4/mod/mod_ssl.html#sslsessiontickets
  {{#if (matchesminorver "2.4" form.serverVersion)}}
    {{#if (minver "2.4.11" form.serverVersion)}}
SSLSessionTickets       off
    {{/if}}
  {{/if}}
{{/if}}
```

generator/src/templates/partials/apache.hbs
```diff
@@ -0,0 +1,59 @@
...
+  {{#if (matchesminorver "2.2" form.serverVersion)}}
+    {{#if (minver "2.2.24" form.serverVersion)}}
+SSLCompression          off
+    {{/if}}
+  {{/if}}
+  {{#if (matchesminorver "2.4" form.serverVersion)}}
```
gene1wood: Or what about a `minpatchver` helper? so you get

```
{{#if (minver "0.9.8" form.opensslVersion)}}
  // https://httpd.apache.org/docs/2.2/mod/mod_ssl.html#sslcompression
  {{#if (minpatchver "2.2" "24" form.serverVersion)}}
SSLCompression          off
  {{/if}}
  // https://httpd.apache.org/docs/2.4/mod/mod_ssl.html#sslcompression
  {{#if (minpatchver "2.4" "3" form.serverVersion)}}
SSLCompression          off
  {{/if}}
{{/if}}
{{#if (minver "0.9.8f" form.opensslVersion)}}
  // https://httpd.apache.org/docs/2.2/mod/mod_ssl.html#sslsessiontickets
  {{#if (minpatchver "2.2" "30" form.serverVersion)}}
SSLSessionTickets       off
  {{/if}}
  // https://httpd.apache.org/docs/2.4/mod/mod_ssl.html#sslsessiontickets
  {{#if (minpatchver "2.4" "11" form.serverVersion)}}
SSLSessionTickets       off
  {{/if}}
{{/if}}
```

generator/src/templates/partials/apache.hbs
```diff
@@ -0,0 +1,59 @@
...
+    {{#if (minver "2.4.3" form.serverVersion)}}
+SSLCompression          off
+    {{/if}}
+  {{/if}}
+{{/if}}
+{{#if (minver "0.9.8f" form.opensslVersion)}}
```
gene1wood: And if you wanted to simply further (in service of making the template more readable), you could add some [logical operator helpers](https://stackoverflow.com/a/31632215/168874) so that you had something like this instead

```
{{#if (and
         (minver "0.9.8" form.opensslVersion)
         (or
           (minpatchver "2.2" "24" form.serverVersion)
           (minpatchver "2.4" "3" form.serverVersion)))}}
SSLCompression          off
{{/if}}
{{#if (and
         (minver "0.9.8f" form.opensslVersion)
         (or
           (minpatchver "2.2" "30" form.serverVersion)
           (minpatchver "2.4" "11" form.serverVersion)))}}
SSLSessionTickets       off
  {{/if}}
{{/if}}
```

generator/src/templates/partials/apache.hbs
```diff
@@ -0,0 +1,59 @@
...
+  {{#if (matchesminorver "2.2" form.serverVersion)}}
+    {{#if (minver "2.2.30" form.serverVersion)}}
+SSLSessionTickets       off
+    {{/if}}
+  {{/if}}
+  {{#if (matchesminorver "2.4" form.serverVersion)}}
```
gene1wood: Or finally you could just create a helper for these two settings to make the template very clear

```
{{#if (apache_ssl_compression_supported form.opensslVersion form.serverVersion)}}
SSLCompression          off
{{/if}}
{{#if (apache_ssl_sessiontickets_supported form.opensslVersion form.serverVersion)}}
SSLSessionTickets       off
{{/if}}
```

generator/src/js/helpers/normalizever.js
april: The library that it uses can only compare semantic versions:
https://semver.org/

Most of the server software generally follows this, but OpenSSL doesn't.

generator/src/templates/partials/apache.hbs
april: That's a pretty good idea, and might clean things up a good deal.

generator/src/templates/partials/apache.hbs
april: I really don't want to write helpers specific to specific pieces of software, it makes it very unclear where the logic is actually happening.

generator/src/templates/partials/apache.hbs
april: I actually find this less readable than what I currently have.  :P

generator/src/js/helpers/normalizever.js
gene1wood: Doesn't line 14 transform things that have only a major and minor version (e.g. `v1.3`) into semantic versions with a major minor and patch, moving the minor version to the patch and setting the new minor version to `0` (e.g. `v1.0.3`)?

dangoncalves: IMHO, you should not rely on semver library.
Instead, you should rely on a String that you will compare (now some projects use Codenames instead of semver).
Then the more powerfull approach may be to use a context and building the configuration regarding this context (e.g.: an apache server with a specific version and OpenSSL with another). I think it may be interesting to work on something like that, but there is a lot of work to do this.
There is an issue about this kind of need: https://github.com/mozilla/server-side-tls/issues/243

gene1wood: I think @dangoncalves is calling out #243 because the Oracle HTTP Server also doesn't use semver (just as openssl doesn't use semver). For example the current version of OHS is [12.2.1.3](https://www.oracle.com/technetwork/middleware/webtier/downloads/index.html)

> you should rely on a String that you will compare 

@dangoncalves do you have a method in mind for how to do version comparisons for these non semver version schemes?

> you should not rely on semver library

wouldn't this still have value for the majority of products that this tool applies to which *do* use semver?

april: Writing string comparisons is a) stupendously complicated and b) prone to breaking. I'd rather deal with the weird one-offs manually by making them into something that semver can understand that write an extremely brittle and complex version comparison library.

april: For reference, the semver library is approximately 1500 lines of code. And while I could certainly do it in fewer lines of code (since there is stuff I don't need), I doubt I could do it nearly as well. And I would be stuck maintaining it in perpetuity.

generator/src/js/helpers/normalizever.js
```diff
@@ -0,0 +1,14 @@
+import { coerce } from 'semver';
+
+export default (version) => {
+  const letterRE = /([abcdefghijlmnopqrstuvwxyz]+$)/;
+
+  // convert to string, just in case
+  version = String(version);
+
+  // get the prerelease tag
+  const prerelease = version.match(letterRE) !== null ? `-${version.match(letterRE)[0]}` : '';
```
gene1wood: I think this may not do what we want.

For openssl it looks like [`1.1.0` is older than `1.1.0a`](https://www.openssl.org/news/changelog.html).

This conversion in `normalizever.js` would turn `1.1.0a` into `1.1.0-a`.

[`1.1.0-a` has a lower precedence than `1.1.0` in semver](https://semver.org/#spec-item-9) which isn't what we want.

These openssl letter suffixes aren't pre-releases, they're "post releases" if that is a thing.

Here's a solution to this
* Take the patch level and multiply it by `100`
* Take the single letter suffix for openssl and convert it to an ordinal number between 1 and 26
* Add the resulting ordinal number to the patch level

This would result in 
* `1.0.0` becoming `1.0.0
* `1.0.0e` becoming `1.0.5`
* `1.0.1` becoming `1.0.100`
* `1.0.2b` becoming `1.0.202`
* `1.1.4k` becoming `1.1.411`

which will make all precedence comparisons work

gene1wood (changes requested):

generator/src/js/helpers/normalizever.js
gene1wood: I'd recommend we
* use semver by default
* intentionally use application specific comparison method for non conforming applications (like openssl)

I realized that I think our general case which is supposed to deal with openssl [has a bug](https://github.com/mozilla/server-side-tls/pull/242#discussion_r280588245) and it may benefit us to make this not general (and openssl specific instead). This way later when we need to compare OHS for example we create an application specific method that normalizes OHS versions (possibly in the math way I [suggest below for openssl](https://github.com/mozilla/server-side-tls/pull/242#discussion_r280588245)

dangoncalves: What about regular expression? 
You can find documentation for javascript [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp)
And a tester [here](https://regex101.com/). If you want the source for this tester, there are [here](https://github.com/firasdib/Regex101/issues)

dangoncalves: My Bad... There is no source for the tester. You can use [this](https://github.com/gskinner/regexr/) instead

dangoncalves: Hey wait... @april you already opened an issue about that: #239

april: Yes, every piece of software would have its own template, but not its own versioning code. That's how we ended up with the mess we have now.

generator/src/js/helpers/normalizever.js
tomato42: > Take the single letter suffix for openssl and convert it to an ordinal number between 1 and 26

note that the suffix doesn't have to be single-letter, during 0.9.8 time it got up to 2 letters

gene1wood:
@april 

What was the resolution of this [issue with using letter suffixes as prerelease in openssl](
https://github.com/mozilla/server-side-tls/pull/242/files/11e0467a42c940ca30c303223118dbeb9e4b82b9#r280588245)?

@tomato42 makes a good point about [double letter versions](https://www.openssl.org/news/vulnerabilities-0.9.8.html).

It looks like the order from oldest to newest in openssl is
* 0.9.8a
* 0.9.8y
* 0.9.8za
* 0.9.8zh
* 0.9.8zy

which means you could use this logic instead

* Take the patch level and multiply it by `100`
* Take the first letter suffix for openssl and convert it to an ordinal number between 1 and 26
* If there exists a second letter suffix, convert it to an ordinal number between 1 and 26
* Add the patch level to the ordinal numbers

This would result in 
* `1.0.0` becoming `1.0.0
* `1.0.0e` becoming `1.0.5`
* `1.0.1` becoming `1.0.100`
* `1.0.2b` becoming `1.0.202`
* `1.1.4k` becoming `1.1.411`
* `1.1.4zc` becoming `1.1.429` (`400` + `26` + `3` = `429`)

april:
That was already fixed by simply comparing strings. If that causes problems down the road I can always make things more complicated but it doesn't seem worthwhile to spend time on something more robust when it doesn't matter for a single configuration file we generate.

april:
I would also add that it's unlikely to ever matter, because:

a) 0.9.8 is no longer supported, and
b) OpenSSL has switched to a more sane versioning system (semantic versioning)

gene1wood:
April explained this to me this morning, there's a code block in `minver.js` that [overrides the `semver` libraries default comparison behavior](https://github.com/mozilla/server-side-tls/pull/242/files#diff-1262c2fc33b386287bf9350fab1454ddR9). As a result, despite `semver` comparing `1.2.3-a` being older than `1.2.3`, the code linked above does the opposite and `1.2.3-a` is newer than `1.2.3`. It does this by detecting if the prerelease is alpha and if so doing a [string comparison](https://github.com/mozilla/server-side-tls/pull/242/files#diff-1262c2fc33b386287bf9350fab1454ddR23). This also accommodates multi character prerelease alpha strings

gene1wood (approved):

april:
Closing this, since it now lives in a different repo.

You can continue to follow this work in https://github.com/mozilla/ssl-config-generator

PR #264, opened 2019-09-20
Add note about Windows Server 2008 R2. Fixes #263.

### LINKED ISSUE #263: Current intermediate breaks near-deprecated IE11 on Windows Server 2008R2.
Apparently there is a very small subset of IE11 users that don't support ECDHE/DHE with RSA, meaning that they have no way to successfully handshake to servers using the current Intermediate configuration.

This includes IE11 on Windows Server 2008R2. The best suites it can negotiate are:

`TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA`
`TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA`

You can see this configuration here:

https://www.ssllabs.com/ssltest/viewClient.html?name=IE&version=11&platform=Win%207&key=36

Windows 2008 R2 is at EOL on January 14th, 2020, which makes me somewhat disinclined to update the configuration to support it. But, we have a few options:

• Add the above cipher suites to Intermediate
• Update the note on Intermediate to indicate that you _must_ use an ECDSA certificate to support these systems
• The above, but also add a note to indicate that you can add `TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA` if you need to support those very old systems and cannot support anything but RSA for your certificates

april:


tomato42 (approved):
LGTM

Issue #265, opened 2019-12-27
Different TLS curve names in the json file


fabian-hk:
https://github.com/mozilla/server-side-tls/blob/392d104f17379cf2aabe45c205a65485b78dffe2/json/server-side-tls-conf-5.0.json#L12
https://github.com/mozilla/server-side-tls/blob/392d104f17379cf2aabe45c205a65485b78dffe2/json/server-side-tls-conf-5.0.json#L42
https://github.com/mozilla/server-side-tls/blob/392d104f17379cf2aabe45c205a65485b78dffe2/json/server-side-tls-conf-5.0.json#L90

Is there a reason that the second curve is defined as "prime256v1" in the modern profile but as "secp256r1" in the intermediate and old profile? I think it would be better if it is defined as "prime256v1" everywhere because then it is also consistent with the [Security/Server Side TLS website](https://wiki.mozilla.org/Security/Server_Side_TLS).

april:
This was fixed in [477459e9ebeb4ccf7e68aaad6c1c5f7c7a44174b](https://github.com/mozilla/ssl-config-generator/commit/477459e9ebeb4ccf7e68aaad6c1c5f7c7a44174b). I'll update the wiki shortly.

Thanks for letting me know!

PR #266, opened 2020-01-02
Update to 5.3, point to ssl-config.mozilla.org

april:


april:
Update the website to reflect #265.

PR #275, opened 2020-07-22
Update certificate lifespan to reflect browser policies

april:


april:
Reviewed with multiple folks on Slack

Issue #247, opened 2019-05-17
Cipher_suites table on wiki may no longer be generated automatically, and therefore stale.


charlesmanser:
https://wiki.mozilla.org/Security/Server_Side_TLS#Cipher_suites states that the table is automatically generated via https://github.com/marumari/tls-table/blob/master/tls-table.py. That returns a 404, not found error.

The table may be stale now. It seems to recommend TLS_DHE_RSA_WITH_AES_128_CBC_SHA256 as intermediate; however, sections like "Intermediate compatibility (default)" and https://statics.tls.security.mozilla.org/server-side-tls-conf.json do not recommend CBC-based ciphers.

floatingatoll:
To address one point only, that username changed a while ago; the corrected
URL would, assuming it’s broken for just the rename and no other reason, be:

https://github.com/april/tls-table/blob/master/tls-table.py

april:
It was always generated manually, and it uses the cipher suites on the wiki page, not the JSON (which I'm not sure existed at the time). It has been updated, but I'm keeping this open until everything is synced.

https://wiki.mozilla.org/Security/Cipher_Suites

charlesmanser:
For example, when I check out https://wiki.mozilla.org/Security/Cipher_Suites, I see green highlighted TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384; however, https://wiki.mozilla.org/Security/Server_Side_TLS#Modern_compatibility does not list any _CBC_/-cbc- ciphers.

janbrasna:
Should be updated now with v5.7 — although manually as seen in #296 

Issues with generating by `tls-table` tracked:
- https://github.com/mozilla/server-side-tls/issues/292
- https://github.com/april/tls-table/issues/4
- https://github.com/april/tls-table/issues/5
- https://github.com/april/tls-table/issues/8

PR #291, opened 2023-05-15
Add TLS_DHE_RSA_WITH_CHACHA20_POLY1305_SHA256 cipher to intermediate configuration

### LINKED ISSUE #285: TLS_DHE_RSA_WITH_CHACHA20_POLY1305_SHA256
The IETF [recommends](https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-4) TLS_DHE_RSA_WITH_CHACHA20_POLY1305_SHA256, but it is not included in the intermediate configuration.

This is causing some people issues.  In particular, those subject to security audits that use different reference material that includes server-side-tls find this to be problematic.

Note that DHE suites in TLS 1.2 are problematic.  [Clients cannot ensure that guarantee that safe groups are used](https://mailarchive.ietf.org/arch/msg/tls/bAOJD281iGc2HuEVq0uUlpYL2Mo/) unless they are willing to retry connections (which comes with its own risks).  The TLS working group is discussing their deprecation, but has not reached a firm recommendation.

My suggestion is that we ignore the controversy until it resolves and restore consistency.  Adding this suite to the intermediate configuration will fix the inconsistency.

gene1wood:
This adds the `TLS_DHE_RSA_WITH_CHACHA20_POLY1305_SHA256` / `DHE-RSA-CHACHA20-POLY1305` cipher to the end of the intermediate cipher list.

See the related PR https://github.com/mozilla/ssl-config-generator/pull/204

Fixes #285

### DIFF
--- a/Cipher_Suites.mediawiki
+++ b/Cipher_Suites.mediawiki
@@ -1,6 +1,9 @@
 = Cipher suite correspondence table =
 IANA, OpenSSL and GnuTLS use different naming for the same ciphers. The table below lists each cipher as well as its corresponding Mozilla [[Security/Server Side TLS|Server Side TLS]] compatibility level.
 
+<!-- Modern row color : #9EDB58 -->
+<!-- Intermediate row color : #DBC158 -->
+<!-- Old row color : #CCCCCC -->
 {| class="wikitable sortable"
 |-
 ! scope="col" | Hex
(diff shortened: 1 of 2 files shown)


Server_Side_TLS.mediawiki
```diff
@@ -202,6 +206,14 @@ This configuration is compatible with a number of very old clients, and should b
 ! Editor
 ! Changes
 |-
+| style="text-align: center;" | 5.7
```
martinthomson, line 209: Up on line 198 and 200, you need to update the JSON file link from 5.6 to 5.7.

Cipher_Suites.mediawiki
```diff
@@ -1,6 +1,9 @@
 = Cipher suite correspondence table =
```
martinthomson, line 1: This table is apparently generated.  I ran the script and it seemed to fail spectacularly, so maybe don't worry about that.  I was hoping that it wouldn't be bitrotten, but I don't think that it's worth fixing right now (there are too many things broken).

martinthomson (approved):
Conditional on updating the links, this LGTM

Cipher_Suites.mediawiki
gene1wood: Oh, good point, I didn't notice the reference to https://github.com/april/tls-table/blob/master/tls-table.py

I'll open an issue about it and we can ignore it for now.

Server_Side_TLS.mediawiki
gene1wood: I fixed this but forgot to commit it so it's in #293

Cipher_Suites.mediawiki
janbrasna: Thanks for updating this manually! It's been stale for the last couple revisions, see #247 

I'll try to update it in a PR to match the current configurations as it is completely outdated just glancing at the colors;)

PR #293, opened 2023-05-16
Fix version number referenced in wiki page

gene1wood:
See #291


janbrasna:
@gene1wood When you have a second please also push a tag and publish a [release](https://github.com/mozilla/server-side-tls/releases) for consistency. Thanks!

Issue #292, opened 2023-05-16
tls-table doesn't seem to be working


gene1wood:
This tool

https://github.com/april/tls-table/blob/master/tls-table.py

is used to produce the Cipher Suites mediawiki page but it doesn't seem to work. We should either get it working or remove reference to it if we won't use it going forward.

janbrasna:
Upstream issues:
- https://github.com/april/tls-table/issues/5
- https://github.com/april/tls-table/issues/8

janbrasna:
I've commented out the reference for now in #296 and made all the necessary updates manually.

In the future I'll look into what changes are needed to parse the new C header file definition formats in `tls-table` but my guess is the table here will need to be maintained manually anyways.

gstrauss:
> This tool
> 
> https://github.com/april/tls-table/blob/master/tls-table.py
> 
> is used to produce the Cipher Suites mediawiki page but it doesn't seem to work. We should either get it working or remove reference to it if we won't use it going forward.

Would you please be more specific about what does not seem to work?  Then, I can take a closer look, and possible choose different data sources.

---

`openssl ciphers -V -stdname` can be used to obtain an easily parseable list of standard cipher names and their openssl names, and can be matched up to the IANA TLS Cipher Suites code points.
```
$ openssl ciphers -V -stdname
          0x13,0x02 - TLS_AES_256_GCM_SHA384                        - TLS_AES_256_GCM_SHA384         TLSv1.3 Kx=any      Au=any   Enc=AESGCM(256)            Mac=AEAD
          0x13,0x03 - TLS_CHACHA20_POLY1305_SHA256                  - TLS_CHACHA20_POLY1305_SHA256   TLSv1.3 Kx=any      Au=any   Enc=CHACHA20/POLY1305(256) Mac=AEAD
          0x13,0x01 - TLS_AES_128_GCM_SHA256                        - TLS_AES_128_GCM_SHA256         TLSv1.3 Kx=any      Au=any   Enc=AESGCM(128)            Mac=AEAD
          0x13,0x04 - TLS_AES_128_CCM_SHA256                        - TLS_AES_128_CCM_SHA256         TLSv1.3 Kx=any      Au=any   Enc=AESCCM(128)            Mac=AEAD
[...]

PR #296, opened 2024-03-14
Update Cipher_Suites table to v5.x

### LINKED ISSUE #247: Cipher_suites table on wiki may no longer be generated automatically, and therefore stale.

janbrasna:
Fixes #247, #292, april/tls-table#4

Since the simplification after v4.0 and the archival of older more expansive docs aligned with v5.0 guidelines, the link to `Cipher_Suites` was lost, and later the table became outdated, with the tool to help gather the cipher lists from source not able to automate the parsing anymore.

Nonetheless the page is still linked from outside, and as a part of the recommendations should be kept up-to-date, even if it means **manually for now**.

There are two major signs the [Cipher_Suites](https://wiki.mozilla.org/Security/Cipher_Suites) page should not be considered abandoned:
1. After archiving the v4 specs to [Security/Archive/Server Side TLS 4.0](https://wiki.mozilla.org/Security/Archive/Server_Side_TLS_4.0) the [Cipher_Suites](https://wiki.mozilla.org/Security/Cipher_Suites) page indeed did become orphaned, but not moved to `*/Archive/*` and with no deprecation warning in the page intro, so there's no indication it should be interpreted as unmaintained.
2. Unlike the archived content that is not included in this repository, the `Cipher_Suites.mediawiki` page is kept here in master meaning it should still be considered part of the current docs, and kept in sync with the guidelines.

This PR makes the table useful again by:
- simplifying some of the styling to be more human-friendly for manual edits (while making the output also more readable/accessible due to better contrast),
- fixing parsing errors for the OpenSSL values used in current configs,
- **updating the color-coding and priority ordering** to reflect the **current guidelines** (v5.7),
- linking the table from main page and correctly linking the configs back, 
- while removing mentions of automated updates (commented out along with timestamps et al. for now) that are no longer true.
- (+ as a bonus fixes reported syntax highlighting mediawiki issue by just using `<pre/>` instead of `<source/>`…)

### This is how the render looks like:

Screen Shot 2024-03-13 at 23 55 37


janbrasna:
The table is also being referenced from MDN e.g. https://developer.mozilla.org/en-US/docs/Web/Security/Transport_Layer_Security via the main wiki page, so the link should be there visible, somewhere.

gstrauss:
> Nonetheless the page is still linked from outside, and as a part of the recommendations should be kept up-to-date, even if it means manually for now.

Agreed.  Additionally, the color coding consistency in #297 is a desirable update.

gstrauss:
> +Different libraries support different cipher suites and refer to them by different names. Mozilla maintains a list of [[Security/Cipher Suites|all known cipher suites]] and their corresponding names.

"all known cipher suites" might be better phrased as "IANA TLS Cipher Suites"
"and their corresponding names" might be better phrased as "and their corresponding names/identifiers used by different libraries".

Overall, the general tone of the changes looks good to me.  I did not verify the manual changes, and will look at #292 to see about fixing the automation.

janbrasna:
The automation for [--colorize](https://github.com/mozilla/server-side-tls/issues/292#issuecomment-2468730911) (that reads the actual mediawiki, to alter the existing content, honoring grading and order) seems pretty broken, and the output mapping even for the basic scenarios including some [gaps](https://github.com/april/tls-table/pull/10#issuecomment-2468742256). At some point April mentioned the script served as the basis for the table in its inception, but was managed mostly manually ever since — not sure if the parsing can process the current content at all.

The wording is taken mostly from an archived version; I'll look into improving that.

Server_Side_TLS.mediawiki
```diff
@@ -73,6 +73,8 @@ Mozilla maintains three recommended configurations for servers using TLS. Pick t
 <p style="max-width: 60em;">The ordering of cipher suites in the <span style="color: gray; font-weight: bold;">Old</span> configuration is very important, as it determines the priority with which algorithms are selected.</p>
 
 <p style="max-width: 60em;">OpenSSL will ignore cipher suites it doesn't understand, so always use the full set of cipher suites below, in their recommended order. The use of the <span style="color: gray; font-weight: bold;">Old</span> configuration with modern versions of OpenSSL may require custom builds with support for deprecated ciphers.</p>
+
+<p style="max-width: 60em;">Different libraries support different cipher suites and refer to them by different names. Mozilla maintains a list of [[Security/Cipher Suites|all known cipher suites]] and their corresponding names.</p>
```
janbrasna, line 77:
```suggestion
<p style="max-width: 60em;">Different libraries support different cipher suites and refer to them by different names. Mozilla maintains a correspondence table of [[Security/Cipher Suites|cipher suites and their mapping]] to identifiers used by different libraries.</p>
```

gene1wood:
I've deployed (copy/pasted) this to wiki.mozilla.org

PR #310, opened 2026-04-29
Remove link to previous, now very old, guidlines

gene1wood:


gene1wood:
This is deployed in https://wiki.mozilla.org/index.php?title=Security/Server_Side_TLS&oldid=1257130

